PR #1, opened 2016-01-06
Hide the hamburger dropdown when no menu specified in opts.

knowuh:
This is for part of [this story](https://www.pivotaltracker.com/story/show/110292104)

For optionally hiding the CFM menu dropdown when embedded in CODAP (in LARA)

[#110292104]


knowuh:
Small change, but I would still appreciate a glance @dougmartin

dougmartin:
Looks good :+1:

PR #2, opened 2016-01-20
Support menu UI in a client-specified element without iframe

kswenson:
Support for clients (such as CODAP) that want to embed the CFM and have the menubar UI rendered in an element specified by the client. The new behavior is specified by passing a new property (appOrMenuElemId). Shouldn't affect the behavior of existing use cases.
- Gulp configuration to allow client to specify alternate build destination.
- Add view-no-frame example demonstrating the menubar being rendered in a client-specified element without an iframe.
- Add appOrMenuElemId property to CFM client configuration.
- Menubar UI renders if appOrMenuElemId property. Client app is only rendered if usingIframe is true.
- Modal dialog code shadows the entire viewport.
- Minor CSS tweaks to explicitly specify styles that might otherwise be overridden by client defaults.

### DIFF
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -3,7 +3,7 @@ var argv = require('yargs').argv,
     production = !!argv.production,
     buildInfo = argv.buildInfo || 'development build (' + (new Date()) + ')',
     src = './src',
-    dest  = './dist';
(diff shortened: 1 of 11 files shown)


sfentress:
This looks good to me. :+1:

PR #3, opened 2016-01-29
Add cfmVersion to CFM metadata

kswenson:
Add cfmVersion to CFM metadata
Use package-json-verify to strip down package.json to just the version.


sfentress:
Looks great.

PR #4, opened 2016-01-29
CODAP-inspired save improvements

kswenson:
When saving to the Concord Document Store:
-- compress POST data
-- only send patches if the patches are smaller than the full document would be
On save failure:
-- clear the "Saving..." message
-- disable autosave until next successful manual save (to prevent accumulation of error dialogs)
Use 'type' instead of 'method' in jQuery ajax calls to support older jQuery versions (like the one used by CODAP/SproutCore 1.9).


sfentress:
This all looks good.

PR #6, opened 2016-02-02
Read only provider supports remote documents

kswenson:
Supports loading of remote documents as well as remote directories containing lists of documents, thus allowing arbitrary sample document libraries.
-- expands the JSON format to support the new features while maintaining compatibility with the old JSON format
-- adds a 'description' property for documents, which is presented as a tooltip
-- adds a 'url' property for documents which is used to load remote documents
-- adds a 'label' type which is used to show dividers and other non-selectable items in file lists
-- adds the 'es6-promise' module for use in handling async requests for remote directories

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "diff": "^2.2.1",
+    "es6-promise": "^3.0.2",
     "jiff": "^0.7.2",
     "jquery": "^2.1.3",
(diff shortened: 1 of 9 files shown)


kswenson:
@dougmartin Sam suggested that you should probably review this as well.

dougmartin:
Looks good.  I live the use of promises

PR #8, opened 2016-02-02
Eliminate double-JSON.stringify on save/patch

kswenson:
-- Eliminates a redundancy introduced in an earlier PR
-- Clears @previouslySavedContent when not in use. (Not technically part of the main purpose of the PR, but a hypothesis I tested while debugging a patch failure that seemed like a good idea.)


dougmartin:
Looks good :+1:

PR #9, opened 2016-02-03
Disable auto-save when a save is pending

kswenson:
I ran into this on the train with four or five auto-save requests pending simultaneously.

### DIFF
--- a/src/code/client.coffee
+++ b/src/code/client.coffee
@@ -191,6 +191,9 @@ class CloudFileManagerClient
             return @_error(err) if err
             @_fileChanged 'openedFile', content, metadata, {openedContent: content.clone()}, @_getHashParams metadata
 
+  isSaveInProgress: ->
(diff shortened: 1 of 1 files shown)


src/code/client.coffee
```diff
@@ -362,11 +365,14 @@ class CloudFileManagerClient
     if @_autoSaveInterval
       clearInterval @_autoSaveInterval
 
+    shouldAutoSave = =>
+      @state.dirty and not @state.metadata?.autoSaveDisabled and not @isSaveInProgress() and @state.metadata?.provider?.can 'save'
+
```
dougmartin: Just a very minor nit: I think if you hoisted shouldAutoSave up into the class definition the setInterval() below would save a tiny amount of memory as the runtime wouldn't need to maintain the closure around autosave().

dougmartin:
Looks good :+1:

PR #10, opened 2016-02-03
Add gulp command-line options (--nojQuery and --codap) for CODAP configuration

kswenson:
Support CODAP gulp configuration via command-line options rather than as a branch or other less desirable alternative.
  --nojQuery flag excludes jQuery for use with clients that already have jQuery installed.
  --codap flag adds '.ignore' to the .js filenames to hide them from the SproutCore build system.
  --codap flag also excludes most assets except a handful of images, since the others are either already present in CODAP (e.g. fonts) or not used in CODAP (e.g. examples).
  --noMap flag isn't CODAP-specific -- it disables inclusion of source maps thus allowing debugging with the browser's JavaScript debugger rather than the CoffeeScript debugger.

Shouldn't affect the behavior of the build unless one of the new options is specified.

### DIFF
--- a/gulp/tasks/browserify-task.js
+++ b/gulp/tasks/browserify-task.js
@@ -4,9 +4,15 @@ var source      = require("vinyl-source-stream");
 var coffeeify   = require('coffeeify');
 var versionify  = require('package-json-versionify');
 var pkgVersion  = require('../../package.json').version;
+var gulpif      = require('gulp-if');
+var rename      = require('gulp-rename');
 var replace     = require('gulp-replace');
 var production  = require('../config').production;
 var config      = require('../config').browserify;
+var flags       = require('../config').flags;
+var noMap       = flags && flags.noMap;
+var nojQuery    = flags && flags.nojQuery;
+var codap       = flags && flags.codap;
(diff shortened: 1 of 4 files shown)


kswenson:
@dougmartin @sfentress I'm not sure who the best reviewer is, so I'll let you decide.

gulp/tasks/browserify-task.js
```diff
@@ -28,18 +34,27 @@ gulp.task('browserify-app', function(){
...
   });
+  if(nojQuery) {
+    b.exclude('jquery');
+  }
   b.transform(coffeeify);
   b.add(config.globals.src);
   return b.bundle()
     .on('error', errorHandler)
     .pipe(source('globals.js'))
+    .pipe(gulpif(nojQuery, replace(/(var)*[ ]*jQuery[ ]*=/g,
```
dougmartin: You should probably change this to 
    /(var)*\s*jQuery\s*=/g
in case someone starts using tabs in the future.

kswenson:
@dougmartin I updated the regex and rebased to master.

dougmartin:
Looks good :+1:

PR #11, opened 2016-02-04
Fix share dialog when used in CODAP/CFM

kswenson:
* default text color to black in share dialog
* add existence test for metadata when handling errors such as failure to save when enabling sharing
* saving for sharing purposes use same compression protocol as simply saving


dougmartin:
Looks good :+1:

PR #12, opened 2016-02-04
Add sourceURL line in CODAP build to simplify debugging

kswenson:
Add sourceURL line in CODAP build to simplify debugging for clients that dynamically load CFM (e.g. CODAP). While in the build system, use the 'extname' option of gulp-rename to simplify file renaming.

@dougmartin Since you reviewed the earlier gulp changes, you may want to take a look at these.


dougmartin:
Looks good :+1:

PR #13, opened 2016-02-10
Fix Google Drive save regression

kswenson:
Oops! In an earlier spate of jQuery 1.6 compatibility changes -- renaming 'method' to 'type' in jQuery ajax calls -- I managed to make the same change in the GoogleDriveProvider, which doesn't use ajax and therefore shouldn't be changed.


dougmartin:
Looks good :+1:

PR #14, opened 2016-02-12
Support opting out of CFM wrapped file format

kswenson:
Adds support for clients to opt out of the CFM wrapped file format.
Add callback to 'openedFile' API for client to indicate successful completion of open.
Clients that opt out of the CFM wrapper file format are responsible for combining CFM shared metadata with the content in 'getContent' and unpacking CFM shared metadata from the content in 'openedFile'.
Use protocol-relative URLs for http/https compatibility.
Add example document demonstrating opting out of the CFM wrapped file format.
Made the other examples consistent with the 'openedFile' API changes by calling back to acknowledge success.
Eliminated a number of JSHint warnings in examples.

@dougmartin Since @sfentress is out of town for a while, you're the likely candidate to review this. I already did a side-by-side review with Sam but this has been modified somewhat since then.

### DIFF
--- a/src/assets/examples/all-providers.html
+++ b/src/assets/examples/all-providers.html
@@ -41,11 +41,11 @@
             help: "http://lmgtfy.com/"
           }
         }
-      }
+      };
       CloudFileManager.createFrame(options, "wrapper", function (event) {
         if (event.type == 'connected') {
           var client = event.data.client;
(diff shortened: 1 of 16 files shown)


sfentress:
This looks fine to me.

@dougmartin since you're currently working on CODAP CFM, when you confirm that this all seems to work as expected do you want to pull it in?

PR #15, opened 2016-03-08
The CFM dialogs should be centered vertically on the browser window [#114585735]

dougmartin:


dougmartin:
@sfentress Simple css fix plus auto re-sizing the dialog content area like we are sizing the gray overlay.

sfentress:
:+1:

PR #16, opened 2016-03-08
Add optional close file dialog for CODAP [#114585439]

dougmartin:


dougmartin:
@sfentress This will be used by CODAP to close the current file.

sfentress:
Can you add to the example app? I just tried it myself, and couldn't tell how to listen for the close event in the app, to clear the content. It only seems to emit a `stateChanged` event.

dougmartin:
@sfentress I've added the example page.

sfentress:
In the example Close doesn't do anything. Shouldn't it clear the page? That's why I was asking about a named event that the app can listen to. But maybe I'm confused as to what Close is supposed to do.

ddamelin:
I think ultimately we want close to clear the document and then show an initial dialog which probably hasn't been made yet, that give the user three choices: start new document, open existing document, browse example documents.

dougmartin:
I've added a 'closedFile' emitted event and handled it in the example app.  I'm going to merge this PR since it touches files I need to work on this morning.

PR #17, opened 2016-03-08
Removed copy link in share dialog and fixed sharing of unsaved content [#113468211]

dougmartin:


dougmartin:
@sfentress The execCommand to copy to info was returning false (even after I fixed the code to use @state.link instead of @props.url) so I'm not sure if this is a change in Chrome's security model (it only allows exec commands on direct user actions) but since this was a feature that I added without it being requested I removed it.  I also fixed a bug I saw when you try to share content that hasn't been saved yet.  The currentContent state variable would be null and the share provider tries to use it.

sfentress:
Hmmm, if I use the original copy code, and change it to use `mark.innerHTML = @state.link`, the copy function seems to work fine for me.

I'm on Chrome 48.0.2564.

I haven't seen anything about the security model changing. That would be weird, as `execCommand` was just added in 42, right?

dougmartin:
@sfentress  You are right - I've updated the code and the commit message.

PR #18, opened 2016-03-08
Added include share data option in download dialog [#114587991]

dougmartin:


dougmartin:
@sfentress Since we keep going back and forth on this I made it an opt-in option to include the share data in the download if the current file is shared.

sfentress:
:+1:

PR #19, opened 2016-03-09
Changed browser alerts to centered alert dialogs [#114585515]

dougmartin:
Adds an alert dialog that can be overlayed on existing dialogs if need be.  A zIndex property was added to the modal view to accomplish this overlay.

### DIFF
--- a/src/assets/examples/menu-madness.html
+++ b/src/assets/examples/menu-madness.html
@@ -42,22 +42,22 @@
         }
       };
 
-      function createMenuItem(text, disabled) {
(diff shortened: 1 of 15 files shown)


dougmartin:
@sfentress Lots of files touched in this one.  For a demo either 1) use the "Menu Madness" example and choose a added menu item to see the alert dialog or 2) enable sharing that then click on the copy link to see the alert overlayed over the share dialog.

dougmartin:
I'm going ahead and merging this as it touches a lot of files I'm working on.

PR #25, opened 2016-04-19
Improved handling of doc store authorization errors

kswenson:
- better detection/handling of expired/stale credentials [#116484877]
- automatically bring up login dialog when possible/appropriate
- dismiss login dialog appropriately upon success [#116342015]
- clear window.location.hash when appropriate

While we're here also, update some dependencies (including React 15.0.1) and add convenience scripts to the package.json.

### DIFF
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -3,10 +3,10 @@ var argv = require('yargs').argv,
     production = !!argv.production,
     buildInfo = argv.buildInfo || 'development build (' + (new Date()) + ')',
     src = './src',
-    dest  = argv.dest ? argv.dest : './dist',
-    noMap = !!argv.noMap,
-    nojQuery = !!argv.nojQuery,
-    codap = !!argv.codap,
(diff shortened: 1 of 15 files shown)


knowuh:
looks good @kswenson 

The only comments I have are food for thought.  

There are several places where the same string literal in English is repeated for authorization errors. Doesn't seem like user-facing messages, but if it were, you could consider using the `t()` functions there?  

I also wonder if some of the boilerplate could be reduced in the `save` / `remove` / `share` / `rename` methods in the future.

👍

kswenson:
Thanks @knowuh. The CFM hasn't generally been using the `t()` functions for error messages (although it should) and so I decided to follow existing practice for consistency. Given that I was only given a day to work on it (although I ended up stretching it to 1.5), I felt it wise not to take on any refactoring efforts that weren't strictly required. Given more time, the ones you mention would be good candidates.

PR #26, opened 2016-05-18
Use run key from LARA when saving as a student

sfentress:
This allows student to launch shared documents from LARA and save back to their own copies.

When a student launches using from LARA using a document on the document-server which was shared by an author, they do not have write-access to that file. Instead, LARA also providers them with the runKey associated with that document in the url as a query parameter.

If we save to the same document using that runKey, then the document-server will automatically create a copy for us, allowing us to save our changes to it. It will then pass down the new id for the file we have ownership to, allowing us to save to that new document from then on.

We were already capturing a using the new id, but two changes had to be introduced:

1. Capture the runKey and use it on the first save
2. Don't try to save via patch on the first save of a session

The second point means that we won't be as efficient in our first save, as we may do a full-save even on a document we already own. Unfortunately, it doesn't seem that there is a way of distinguishing between a document we own and one we don't, until we have a successful first save. At which point, the record id returned by the server will always be one that we own.

### DIFF
--- a/src/code/utils/get-query-param.coffee
+++ b/src/code/utils/get-query-param.coffee
@@ -1,0 +1,9 @@
+module.exports = (param) ->
+  param = param.replace(/[\[]/, "\\[").replace(/[\]]/, "\\]")
+  regexS = "[\\?&]" + param + "=([^&#]*)"
+  regex = new RegExp regexS
+  results = regex.exec window.location.href
+  if results?.length > 1
+    return decodeURIComponent results[1]
+  else
+    return null
--- a/src/code/app.coffee
+++ b/src/code/app.coffee
@@ -4,6 +4,7 @@ CloudFileManagerUIMenu = (require './ui').CloudFileManagerUIMenu
 CloudFileManagerClient = (require './client').CloudFileManagerClient
 
 getHashParam = require './utils/get-hash-param'
+getQueryParam = require './utils/get-query-param'
 
 class CloudFileManager
 
@@ -24,6 +25,7 @@ class CloudFileManager
       sharedContentId: getHashParam "shared"
       fileParams: getHashParam "file"
       copyParams: getHashParam "copy"
+      runKey: getQueryParam "runKey"
     }
 
     @client.setAppOptions @appOptions
--- a/src/code/providers/document-store-provider.coffee
+++ b/src/code/providers/document-store-provider.coffee
@@ -198,9 +198,6 @@ class DocumentStoreProvider extends ProviderInterface
       xhrFields:
         {withCredentials}
       success: (data) ->
-        # record previously saved content for patching purposes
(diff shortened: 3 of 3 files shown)


sfentress:
@kswenson can you take a look at this when you have a chance?

src/code/utils/get-query-param.coffee
```diff
@@ -0,0 +1,9 @@
+module.exports = (param) ->
+  param = param.replace(/[\[]/, "\\[").replace(/[\]]/, "\\]")
+  regexS = "[\\?&]" + param + "=([^&#]*)"
+  regex = new RegExp regexS
+  results = regex.exec window.location.href
+  if results?.length > 1
+    return decodeURIComponent results[1]
+  else
+    return null
```
kswenson: Couldn't this be simplified to `results?.length > 1 ? decodeURIComponent results[1] : null`?

src/code/utils/get-query-param.coffee
```diff
@@ -0,0 +1,9 @@
+module.exports = (param) ->
+  param = param.replace(/[\[]/, "\\[").replace(/[\]]/, "\\]")
```
kswenson: Perhaps a comment as to the handling of '[' and ']'?

kswenson:
Looks good! The runKey handling seemed complicated when Jonathan and I were trying to figure it out but seems remarkably simple when presented/handled this way. ;-) You may merge at will. I added a couple of line comments on the implementation of getQueryParam(), but you may take them or leave them as you see fit.

ddamelin:
We should make sure to try creating a doc with CFM CODAP, getting the LARA link from the Share dialog, putting that in a LARA doc, and making sure everything works as expected from the LARA user perspective. This is slightly different because of the way ownership is handled with the new way we share.

src/code/utils/get-query-param.coffee
sfentress: The day that coffeescript supports ternary operators will be a day that I rejoice! Until then...

(In any case, even if it supported it, it's just a style question, not a logic simplification. And I actually like having `return` statements be fairly explicit, particularly in if-blocks, even when CS returns things implicitly.)

kswenson: Right, I'm thinking JS. In CS it would be `if results?.length > 1 then decodeURIComponent results[1] else null` but as you say, a matter of preference.

src/code/utils/get-query-param.coffee
sfentress: Well, I could do, but then I'd actually have to spend more time understanding the regex, instead of grabbing this uncommented code that someone committed to CODAP... https://github.com/concord-consortium/codap/blob/0b302ae3927ac2988dfc2cdcd8261202bec58716/apps/dg/core.js#L128-139

😜 😜

PR #27, opened 2016-05-31
Improve runKey handling

kswenson:
- If runKey was passed as URL parameter, pass it to document server when loading or saving the document.
This is necessary to support use cases such as teachers viewing student work.
- Support opening and saving documents without authorization when 'runAsGuest' parameter is specified
- Fix copy/paste errors in some request error handlers


kswenson:
Merging since CODAP has been running with this code for over a month.

kswenson:
Merging since CODAP has been running with this code for over a month and Jonathan has some followup commits he'd like to make that depend on these changes.

PR #28, opened 2016-07-07
Fix download and upload

jsandoe:
These are three small bug fixes.

  * Couldn't download large documents. This was because a data uri was used. Converted to use a Blob uri. PT [#121523735]
  * The "all changes saved" message was no longer displaying when a save completed. The saved state was being overwritten inadvertently. PT [#117561145]
  * Finally a drop of a uri list was not correctly handled.

### DIFF
--- a/src/code/views/download-dialog-view.coffee
+++ b/src/code/views/download-dialog-view.coffee
@@ -35,8 +35,10 @@ module.exports = React.createClass
 
   download: (e) ->
     makeBlobURL = (msg) ->
+      wURL = window.URL or window.webkitURL
       blob = new Blob([msg], {type: 'text/plain'})
-      window.URL.createObjectURL(blob)
+      if (wURL)
+        wURL.createObjectURL(blob)
 
     if @state.trimmedFilename.length > 0
       json = @props.content.getContent()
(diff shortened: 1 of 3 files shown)


src/code/views/download-dialog-view.coffee
```diff
@@ -34,6 +34,10 @@ module.exports = React.createClass
     s.replace /^\s+|\s+$/, ''
 
   download: (e) ->
+    makeBlobURL = (msg) ->
+      blob = new Blob([msg], {type: 'text/plain'})
+      window.URL.createObjectURL(blob)
```
kswenson: I should have asked about browser compatibility when we were looking at this together. Looking at the Browser compatibility chart at https://developer.mozilla.org/en-US/docs/Web/API/Window/URL, it browser compatibility can be improved by changing this to something like:
```
wURL = window.URL || window.webkitURL
if (wURL) wURL.createObjectURL(blob)
```

jsandoe: Made the suggested change (BTW, the site caniuse.org is also pretty good for compatibility issues. e.g.:  http://caniuse.com/#feat=bloburls)

PR #29, opened 2016-07-25
Remove unused jQueryUI and touchpunch.js

kswenson:
CODAP is (now) using the latest jQueryUI (1.12). CFM was including an old version of jQueryUI (1.10). Given that CFM's UI is React-based and its only use of jQuery is for things like $.ajax() and $.bind(), it seemed likely that CFM didn't use/need jQueryUI at all, which appears to be the case.

Looking at the git history, @dougmartin [added jQuery and jQueryUI](https://github.com/concord-consortium/cloud-file-manager/commit/db8f15a70e8e1e777309a86aee1b483aeeacc653) originally, then [removed jQueryUI](https://github.com/concord-consortium/cloud-file-manager/commit/5dae7e4e19b2c15a27785e0ea443683bc1a9ec35) somewhat later, only to have @knowuh [add it back to fix a failing build](https://github.com/concord-consortium/cloud-file-manager/commit/3b93de93903f117a11dd88c3e4c21605b12d2d4f).

Removing touchpunch.js as well as jQueryUI allows us to avoid the build issue that was previously encountered. @dougmartin I think you're the logical reviewer, particularly in Sam's absence.


dougmartin:
Looks good :+1:

PR #30, opened 2016-08-12
Notify when ready

sfentress:
This adds a client event, `ready`, which is called

1. After the client has connected, if there is nothing left to do (no file to open)
2. After a file has been opened and the app has called the callback

This can be used by the app to know that we're in a ready state and the user can start interacting with the app (hide splash screen, etc.)

Previously, it was not clear from the app perspective whether a file was going to be loaded by the CFM after the initial `connect` event.


kswenson:
Looks good :thumbsup:

PR #31, opened 2016-08-17
Allow readonly files to be loaded from url hash

sfentress:
This gives the readOnly provider the capability already found in the other providers, that of modifying the url hash when a file is opened, and loading a file when a url with a file fragment is opened.

This way we can link directly to a static example, using `...#file=examples:some-page`

Additionally, apps can chose what to name the providers in the url hash, and have multiple named providers so the fragment can be   `...#file=demos:some-page`, `...#file=AP-curriculum:some-page` or whatever makes most sense given the language of the project.

### DIFF
--- a/src/code/providers/readonly-provider.coffee
+++ b/src/code/providers/readonly-provider.coffee
@@ -57,6 +57,8 @@ class ReadOnlyProvider extends ProviderInterface
       # clone the metadata items so that any changes made to the filename or content in the edit is not cached
       callback null, _.map items, (metadataItem) -> new CloudMetadata metadataItem
 
+  canOpenSaved: -> true
+
   openSaved: (openSavedParams, callback) ->
     metadata = new CloudMetadata
       name: unescape(openSavedParams)
--- a/src/code/providers/readonly-provider.coffee
+++ b/src/code/providers/readonly-provider.coffee
@@ -25,24 +25,22 @@ class ReadOnlyProvider extends ProviderInterface
   @Name: 'readOnly'
 
   load: (metadata, callback) ->
-    @_loadTree (err, tree) ->
-      return callback err if err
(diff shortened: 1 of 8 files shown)


sfentress:
This has been deployed to http://concord-consortium.github.io/cloud-file-manager/examples/

Here's an example that opens a remote file, nested in a folder, found in one of four readOnly providers: http://concord-consortium.github.io/cloud-file-manager/examples/read-only-with-folders.html#file=readOnly-remote2:Map%20Data

src/code/providers/readonly-provider.coffee
```diff
@@ -51,7 +57,17 @@ class ReadOnlyProvider extends ProviderInterface
       # clone the metadata items so that any changes made to the filename or content in the edit is not cached
       callback null, _.map items, (metadataItem) -> new CloudMetadata metadataItem
 
-  canOpenSaved: -> false
```
dougmartin: I think this needs to stay and be changed to ```-> true```.  The getHashParams method in the client uses is like ```if metadata?.provider?.canOpenSaved()```.  Maybe that should change to ```if metadata?.provider?.canOpenSaved?()```?

dougmartin:
Other than the question about the removal of canOpenSaved this looks good.

src/code/providers/readonly-provider.coffee
sfentress: It's not strictly necessary, because the ProviderInterface by default returns `true`, so, since we extend it, we return true as well.

I didn't edit the ProviderInterface (it's always been that way, I guess), but it's true that it's ambiguous. Especially since the default `getOpenSavedParams` in ProviderInterface is `notImplemented`.

I think I should change the default to return `false`, and add this back here.

PR #32, opened 2016-08-17
Add ability to open files at arbitrary urls directly

sfentress:
This adds a new provider to load files from remote urls directly.

The app will automatically try to open urls when it sees the url fragment

`#file=http...`

The app can also ask the client to open a url directly through the API, and a CODAP-friendly example of this is added, showing an app that takes a query parameter and uses it to open a remote file.

[#121532705]

### DIFF
--- a/src/code/client.coffee
+++ b/src/code/client.coffee
@@ -8,6 +8,7 @@ ReadOnlyProvider = require './providers/readonly-provider'
 GoogleDriveProvider = require './providers/google-drive-provider'
 DocumentStoreProvider = require './providers/document-store-provider'
 LocalFileProvider = require './providers/local-file-provider'
+URLProvider = require './providers/url-provider'
 
 cloudContentFactory = (require './providers/provider-interface').cloudContentFactory
 CloudContent = (require './providers/provider-interface').CloudContent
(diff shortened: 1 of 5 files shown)


src/code/client.coffee
```diff
@@ -229,6 +230,12 @@ class CloudFileManagerClient
             return @alert(err) if err
             @_fileOpened content, metadata, {openedContent: content.clone()}, @_getHashParams metadata
 
+  openUrlFile: (url) ->
+    urlProvider = new URLProvider()
```
dougmartin: Instead of instantiating it here for each call you could add it to the list of providers that are setup at startup on line 38.

dougmartin:
Other than the question of when to instantiate the url provider this looks good.

src/code/client.coffee
sfentress: It's entirely a question of style. Since we only ever use this at most once (since we don't watch hash changes, to change the file we have to reload the app), and may not use it at all, it seemed to me to make sense to not instantiate it at all unless we need it.

But I can see it either way. Thoughts?

dougmartin: I'd keep it as is but maybe add a comment with why it isn't instantiated like the rest of them.

sfentress: I realized it could be used again, if the app were calling `openUrlFile` from the API. I just instantiated it at the top. (It's not part of `@allProviders` because that's the app-selected list, and we want this to always be there.)

PR #33, opened 2016-08-23
Refactor for v2 API

kswenson:
Implement DocumentStoreUrl class
- Move responsibility for knowing/manipulating document store URLs and parameters into its own class
- Support redirecting document store requests via 'documentServer' command line argument
- Fixes [#128108157]

Separate sharing functionality into DocumentStoreShareProvider class

PR #34, opened 2016-08-24
Rationalize url param handling; improve error handling

kswenson:
(Note: Depends on [PR #33](https://github.com/concord-consortium/cloud-file-manager/pull/33))

In further preparation for the V2 API...

Rationalize URL/hash parameter handling
- hash params are (still) handled by the CloudFileManager (app)
- url params are provider-specific, and so their handling is moved to individual providers (mainly the DocumentStoreProvider at this point)
- hash/url param processing is moved from CloudFileManager.clientConnect() to CloudFileManagerClient.processUrlParams() (which is called from CloudFileManager.clientConnect())
- CloudFileManagerClient.processUrlParams() gives providers a chance to handle hash/url params
- DocumentStoreProvider handles additional recordid url parameter eliminating the need for CODAP to translate it for CFM

Error handling improvements
- allow splash screen to close if an error occurs when opening a document initially (discovered while testing the changes in the previous commits)
- request authorization on failure to open due to lack of authorization
- improve error message upon failure to load from a URL
- signal `ready` on failure to open a file

### DIFF
--- a/src/code/client.coffee
+++ b/src/code/client.coffee
@@ -190,7 +190,7 @@ class CloudFileManagerClient
   openFile: (metadata, callback = null) ->
     if metadata?.provider?.can 'load'
       metadata.provider.load metadata, (err, content) =>
-        return @alert(err) if err
+        return @alert(err, => @ready()) if err
         # should wait to close current file until client signals open is complete
         @_closeCurrentFile()
         @_fileOpened content, metadata, {openedContent: content.clone()}, @_getHashParams metadata
@@ -249,7 +249,7 @@ class CloudFileManagerClient
 
   openSharedContent: (id) ->
     @state.shareProvider?.loadSharedContent id, (err, content, metadata) =>
-      return @alert(err) if err
+      return @alert(err, => @ready()) if err
       @_fileOpened content, metadata, {overwritable: false, openedContent: content.clone()}
 
   # must be called as a result of user action (e.g. click) to avoid popup blockers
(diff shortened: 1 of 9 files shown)


src/code/client.coffee
```diff
@@ -103,6 +103,38 @@ class CloudFileManagerClient
...
+    else if hashParams.copyParams
+      @openCopiedFile hashParams.copyParams
+    else
+      # give providers a chance to process url params
+      for provider in @state.availableProviders
+        handleResult = provider.canHandleUrlParams()
+        if handleResult and handleResult.providerName
+          {providerName, providerParams} = handleResult
+          @openProviderFile providerName, providerParams
+          return
```
sfentress: `canHandleUrlParams` sounds like a generic ability that the provider can choose to act on any arbitrary urlParams, but here it seems to be used simply as "can we call `openProviderFile` with some value?"

Couldn't the provider itself decide what to do with those params? It could call back to `client. openProviderFile`, or it could do something else entirely. Then you could even put the share params in the DocumentServer's `canHandleUrlParams`. (Maybe it would be renamed to `handleUrlParams`.)

kswenson: My thinking was that the preexisting code assumed that the result of processing the hash parameters was a call to openProviderFile; this just extended the same assumption to support resolving the query parameters before a call to openProviderFile. But I tried it out the other way and the code got simpler, so I've added a commit which refactors `canHandleUrlParams` to `handleUrlParams` with the semantics you suggest.

sfentress:
Looks good to me 👍

PR #35, opened 2016-08-30
Let Google Drive create new files in specific folders

sfentress:
This adds a new url hash param, `newInFolder={provider}:{folder-id}`, which signals that the app should initialize a blank doc and automatically save it back to the provider, under the correct folder.

This is used by the Google Drive API to allow a user to click on the New button within a folder, and have the file be saved back there. (This also works from the home folder, naturally.)

When the doc is first opened, the filename input at the top is selected and focused, allowing the user to immediately name their document.

### DIFF
--- a/src/code/app.coffee
+++ b/src/code/app.coffee
@@ -25,6 +25,7 @@ class CloudFileManager
       sharedContentId: getHashParam "shared"
       fileParams: getHashParam "file"
       copyParams: getHashParam "copy"
+      newInFolderParams: getHashParam "newInFolder"
       runKey: getQueryParam "runKey"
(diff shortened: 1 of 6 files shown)


kswenson:
:thumbsup: Looks good to me. There will be a conflict with my URL rationalization PR but it should be easy to address. If we merge this one first I can rebase mine.

kswenson:
With 20/20 hindsight, I should have merged Sam's PRs first. Instead I started with mine (since it was the oldest) and now have some manual merging to do. On the plus side, I'm sure I'll learn something about how merging PRs with conflicts works on Github. :wink:

PR #36, opened 2016-08-31
Support v2 api; add LaraProvider to support documents launched from LARA

kswenson:
Note: Depends on [PR #34](https://github.com/concord-consortium/cloud-file-manager/pull/34)

Support V2 Document Server API

DocStoreShareProvider uses V2 API for sharing and updating shared documents

Add LaraProvider to support documents launched from LARA using the V2 document server API [#128108423]
- LaraProvider base64.decodes the launchFromLara query param to retrieve providerData
- If providerData contains recordID and accessKeys, then the document is loaded directly
- If providerData contains a url, the interactive run state is queried to determine document ID and access keys
- Separate PatchableContent class from DocumentServerProvider to be used by LaraProvider as well
- If a provider returns null from filterTabComponent(), the ProviderTabbedDialog doesn't show it in the dialog


src/code/utils/translate.coffee
```diff
@@ -5,7 +5,8 @@ varRegExp = /%\{\s*([^}\s]*)\s*\}/g
 
 translate = (key, vars={}, lang=defaultLang) ->
   translation = translations[lang]?[key] or key
-  translation.replace varRegExp, (match, key) ->
-    if vars.hasOwnProperty key then vars[key] else "'** UKNOWN KEY: #{key} **"
+  if translation
```
sfentress: Translate can only be falsy if key itself is falsy. Is that likely? We're always passing something into `tr`.

src/code/client.coffee
```diff
@@ -547,6 +547,25 @@ class CloudFileManagerClient
     # Check browser support for document.location.origin (& window.location.origin)
     "#{document.location.origin}#{document.location.pathname}#{suffix}"
 
+  # Takes an array of strings representing url parameters to be removed from the URL.
+  # Removes the specified parameters from the URL and then uses the history API's
+  # pushState() method to update the URL without reloading the page.
+  # Adapted from http://stackoverflow.com/a/11654436.
+  removeQueryParams: (params) ->
```
sfentress: It looks like you're supposed to be able to pass this function an array of query params to eliminate. However, as far as I can tell, it will only ever remove the last element of that array.

`hash` is only calculated once, at the start of the function, and `replace` is called on `hash[0]`, which isn't changing.

A quick fix might be on line 562: `url = hash[0] = hash[0].replace(...`, but there may be a more elegant fix.

kswenson: Good catch. Fixed in a subsequent commit.

src/code/utils/translate.coffee
kswenson: I actually encountered this at some point during development. I don't recall the specific circumstances but it was probably something like the client was using an ill-formed if expression which resulted in passing undefined as the key. It didn't seem appropriate for the translate function to crash at that point, although there may be a better return value in that scenario than the current undefined.

src/code/providers/lara-provider.coffee
```diff
@@ -279,8 +279,7 @@ class LaraProvider extends ProviderInterface
     callback "Cannot open the specified document"
 
   getOpenSavedParams: (metadata) ->
-    @encodeParams if metadata.providerData.url \
-                    then { url: metadata.providerData.url } \
-                    else metadata.providerData
+    # if loaded from lara then reuse the params so we don't leak the keys
+    @encodeParams @laraParams or metadata.providerData
```
kswenson: My assumption here was that after the initial open the run state would be updated so that other fields of the `launchWithLara` params would not be required (or desired). At the time, I was specifically thinking of the `source` parameter, which is clearly not needed after the initial open. Is there a scenario where the access keys need to be in the URL because they're not saved in the run state? Wouldn't that mean that anyone that got hold of the URL could potentially open and/or edit the document, which was what we were trying to get away from by storing the access keys in the run state?

kswenson:
I had one question about the handling of the 'launchWithLara' params in the last commit. Otherwise, everything looks good. 👍

dougmartin:
I removed the metadata fallback in the param encoding in the lara provider

PR #37, opened 2016-08-31
Extensions and mime types

sfentress:
This improves current extension and mimeType handing by

1. Allowing option arrays of additional extensions/mimeTypes to filter on (which should allow us to re-incorporate the extension option in CODAP, as we can now include legacy formats)
2. Not overwriting existing extensions when we open a readable file from a remote location
3. Adding the main extension to downloaded files, but not duplicating it when it is already there


kswenson:
:thumbsup: Looks good to me.

PR #39, opened 2016-09-13
Save secondary file

sfentress:
Allows CFM to save a file to a provider, without affecting the current metadata. Used for exporting .csv files in CODAP.

https://www.pivotaltracker.com/story/show/127512941

### DIFF
--- a/src/code/providers/provider-interface.coffee
+++ b/src/code/providers/provider-interface.coffee
@@ -8,7 +8,7 @@ class CloudFile
 
 class CloudMetadata
   constructor: (options) ->
-    {@name, @type, @description, @content, @url, @provider = null, @parent = null, @providerData={}, @overwritable, @sharedContentId, @sharedContentSecretKey} = options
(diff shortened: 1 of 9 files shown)


src/code/providers/provider-interface.coffee
```diff
@@ -30,6 +30,11 @@ class CloudMetadata
     else
       name
 
+  @newExtension: (name, extension) ->
+    if ~name.indexOf(".")
+      name = name.split(".")[0]
+    name + "." + extension
```
kswenson: In the case of multiple dots in a name, I believe it is customary to assume that the last one indicates the extension rather than the first. A comment under this StackOverflow answer (http://stackoverflow.com/a/4250417) suggests `x.substr(0, x.lastIndexOf('.')) || x`, for instance.

kswenson:
I had one comment about the file extension parsing. Otherwise it looks good. 👍

src/code/providers/provider-interface.coffee
sfentress: Sounds good.

sfentress:
That's now updated.

PR #45, opened 2016-09-23
Fix Save As bug detected during Concord Cloud deprecation testing

kswenson:
When switching providers (e.g. saving a file previously saved to Concord Cloud to Google Drive) the provider-specific data in the metadata must be cleared or it can confuse the new provider.


dougmartin:
Looks good 👍

PR #46, opened 2016-09-29
Added missing check for error in list for Google Drive provider

dougmartin:
This was the only place we were not checking for the error message in the returned results.  This should surface the errors thrown for these two stories:

- https://www.pivotaltracker.com/n/projects/1055240/stories/131230747
- https://www.pivotaltracker.com/n/projects/1055240/stories/131230545


dougmartin:
@kswenson Could you review and update your private build?

kswenson:
Not necessary for this PR, but it looks like the remove and rename methods don't call `@_apiError` on failure either.

PR #47, opened 2016-09-29
Added enableDeflate flags to docstore and lara providers (set to false for now)

dougmartin:
This change is for testing if the client side deflate is causing the patch error in https://www.pivotaltracker.com/n/projects/1055240/stories/131313927

### DIFF
--- a/src/code/providers/document-store-provider.coffee
+++ b/src/code/providers/document-store-provider.coffee
@@ -286,17 +286,21 @@ class DocumentStoreProvider extends ProviderInterface
     if @urlParams.runKey
       params.runKey = @urlParams.runKey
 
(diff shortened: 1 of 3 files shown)


dougmartin:
@kswenson This PR should not be merged but rather just included in your build so that Evangeline can test it.  It looks like the pako deflate is either sending badly encoded payloads or the middleware for the DocStore is not inflating the payload correctly - the patch json is invalid json on the other end.

kswenson:
The compression has been (and probably still is) working fine using the V1 API, so the problem appears to be specific to the V2 API code paths.

kswenson:
Closing now that compression is properly supported server-side.

PR #48, opened 2016-09-29
Add back code to show docstore tab after it is disabled once

dougmartin:


dougmartin:
@kswenson I added this back as I believe the bugfix you added to update the file provider metadata fixes the root bug that I was trying to mask when this was removed.

PR #49, opened 2016-10-01
Fix crash opening LocalFile then saving

kswenson:
Fix crash that occurred when opening a file using the LocalFile provider tab in the open dialog and then saving it
- the FileDialogTab component was calling _.cloneDeep() on the metadata, which contains a pointer back to the provider which is not prepared to be copied (it has a pointer back to the client, etc.)


dougmartin:
Looks good :+1:

PR #51, opened 2016-10-03
If patch save fails, try a full save before reporting error to user

kswenson:
If patch save fails, try a full save before reporting error to user [#130829875]

Note: I started this initially as a means of giving Evangeline an alternative build to test while investigating https://www.pivotaltracker.com/story/show/131313927. It can be merged (or not) at whatever point makes sense relative to that more important issue.


dougmartin:
Looks good 👍

PR #61, opened 2016-11-02
Converted "Download" to "Save" to local file system [#133108353]

dougmartin:
* Added file save to local file provider
* Fixed missing sharing code changes for v2 document store
* Added download/save when enter key is pressed in save and download dialogs
* Removed "Download" from default menu

Viewable here: https://codap.concord.org/~dmartin/add-local-file-save/static/dg/en/cert/index.html

PR #62, opened 2016-11-04
Support saving secondary files (e.g. CSV) to the Local File provider

kswenson:
- allow Local File provider to be selected in export dialog
- support saving/downloading secondary content

This PR ties together Sam's earlier work to [support CSV export](https://github.com/concord-consortium/codap/pull/91) and Doug's recent work to support [local file saving/downloading](https://github.com/concord-consortium/cloud-file-manager/pull/61) so that it is possible to download/save exported CSVs locally. @dougmartin I embarked on this thinking it might be a quick fix before merging your CFM changes into Sam's outstanding PR. It turned out to be a bit more involved.

### DIFF
--- a/src/code/client.coffee
+++ b/src/code/client.coffee
@@ -487,10 +487,14 @@ class CloudFileManagerClient
       @_ui.downloadDialog @state.metadata?.name, envelopedContent, callback
 
   getDownloadUrl: (content, includeShareInfo) ->
-    contentToSave = content
+    if typeof content is "string"
+      contentToSave = content
 
-    if not includeShareInfo
-      # clone the document so we can delete the share info if needed and not effect the original
+    else if includeShareInfo
+      contentToSave = JSON.stringify(content.getContent())
+
+    else # not includeShareInfo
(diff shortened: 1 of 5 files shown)


src/code/client.coffee
```diff
@@ -487,19 +487,22 @@ class CloudFileManagerClient
       @_ui.downloadDialog @state.metadata?.name, envelopedContent, callback
 
   getDownloadUrl: (content, includeShareInfo) ->
-    # clone the document so we can delete the share info if needed and not effect the original
-    json = content.clone().getContent()
+    contentToSave = content
```
dougmartin: Originally the content parameter was an object but you are now passing it as a string when saving secondary files.  I think we should add a type check like:

```contentToSave = if typeof content is 'string' then content else JSON.stringify(content.getContent())```

dougmartin:
Looks good except for the one issue with setting contentToSave in client.coffee

src/code/client.coffee
kswenson: Good catch; fixed.

PR #63, opened 2016-11-05
Support saving exported picture as local file

kswenson:
One more extension to support saving exported graph pictures as local files

### DIFF
--- a/src/code/client.coffee
+++ b/src/code/client.coffee
@@ -487,7 +487,7 @@ class CloudFileManagerClient
       @state.currentContent?.copyMetadataTo envelopedContent
       @_ui.downloadDialog @state.metadata?.name, envelopedContent, callback
 
-  getDownloadUrl: (content, includeShareInfo, mimeType) ->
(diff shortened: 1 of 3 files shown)


src/code/client.coffee
```diff
@@ -486,9 +487,12 @@ class CloudFileManagerClient
       @state.currentContent?.copyMetadataTo envelopedContent
       @_ui.downloadDialog @state.metadata?.name, envelopedContent, callback
 
-  getDownloadUrl: (content, includeShareInfo) ->
+  getDownloadUrl: (content, includeShareInfo, mimeType) ->
```
dougmartin: Since you test mimeType on line 492 (`if mimeType.indexOf("image") >= 0`) and it is an optional parameter I'd set the default value here:

`, mimeType = 'text/plain') ->`

and get rid of the `mimeType or 'text/plain'` on line 513.

dougmartin:
Looks good once you set the mimeType default value.

src/code/client.coffee
kswenson: Good point. Made the suggested change.

PR #64, opened 2016-11-10
Support legacy sharing with 'doc' and 'owner' query parameters

kswenson:
- save triggers save as dialog for saving unshared copy
- fixed bug that could prevent saving to another provider after saving to the local file provider

This was the last bit of document save/restore code still being handled by CODAP. Moving it to the CFM enables a much more complete cleanup of the document save/restore code in CODAP.


dougmartin:
Looks good 👍

PR #65, opened 2016-11-18
Fix save local file to google drive

kswenson:
Files opened from Local can now be saved to Google Drive [#134090113]
- replace 'forceSaveDialog' metadata property with 'resave' provider capability
- rename 'saveUnwrapped' provider capability to 'export'
- update capability specifications for all providers

Fix handling of client-provided submenus


dougmartin:
Looks good 👍

PR #66, opened 2016-11-19
Prevent redundant save dialogs

kswenson:
Once more unto the breach...

With the changes in PR#65, some code paths resulted in the presentation of multiple save dialogs. This occurred because in `CloudFileManagerClient` the `saveFileDialog()` and `saveFileAsDialog()` methods call `_dialogSave()` once the dialog is dismissed. The `_dialogSave()` method calls `saveFile()`, which optionally brings up the save dialog. The changes in PR#65 triggered this latent issue. The fix here is to pull out the actual saving code from `saveFile()` into a `saveFileNoDialog()` method. With this change, `saveFile()` calls `saveFileDialog()` or `saveFileNoDialog()` as appropriate, and `_dialogSave()` calls `saveFileNoDialog()`, thus eliminating the potential for redundant dialogs.

### DIFF
--- a/src/code/client.coffee
+++ b/src/code/client.coffee
@@ -337,26 +337,29 @@ class CloudFileManagerClient
   saveFile: (stringContent, metadata, callback = null) ->
     # must be able to 'resave' to save silently, i.e. without save dialog
     if metadata?.provider?.can('resave', metadata)
-      @_setState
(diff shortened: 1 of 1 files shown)


dougmartin:
Looks good 👍

PR #69, opened 2016-11-30
Misc bug fixes

dougmartin:


src/code/utils/lang/en-us.coffee
```diff
@@ -66,6 +66,7 @@ module.exports =
   "~SHARE_DIALOG.COPY_SUCCESS": "The info has been copied to the clipboard."
   "~SHARE_DIALOG.COPY_ERROR": "Sorry, the info was not able to be copied to the clipboard."
   "~SHARE_DIALOG.COPY_TITLE": "Copy Result"
+  "~SHARE_DIALOG.LONGEVITY_WARNING": "Please note: If this copy of your document is not accessed for more than one year it will be removed."
```
kswenson: I think it would be friendlier to state this in the affirmative, perhaps something like "The shared copy of this document will be retained until it has not been accessed for over a year." Precise wording open for discussion.

bfinzer:
I like Kirk’s amended version of the longevity warning.

ddamelin:
I don't think we need to automatically remove shared files after some time. They typically take up a small amount of space, someone might go a long time between uses, and it adds text to read when sharing which could better be spent reading the important message we already have there about this being a copy of the current state. I would not implement this feature.

bfinzer:
Actually shared files can be arbitrarily large. Without this warning and mechanism we have no way to clear out files in the future. And the user has no understanding of whether the shared file is likely to disappear in a week or will be there permanently.

dougmartin:
Here is the updated message (again not enabled by default):

image

dougmartin:
@ddamelin @bfinzer Should I merge this in now with the updated text?  The warning is not enabled and the text can always be tweaked later when a policy is decided upon.

bfinzer:
Yes, please.

PR #70, opened 2016-11-30
Added docstore call when removing sharing [#130691121]

dougmartin:
NOTE: this is a noop until this docstore PR is merged and deployed:  https://github.com/concord-consortium/document-store/pull/21

### DIFF
--- a/src/code/providers/document-store-share-provider.coffee
+++ b/src/code/providers/document-store-share-provider.coffee
@@ -24,7 +24,7 @@ class DocumentStoreShareProvider
   getSharingMetadata: (shared) ->
     { _permissions: if shared then 1 else 0 }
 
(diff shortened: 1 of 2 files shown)


src/code/providers/document-store-share-provider.coffee
```diff
@@ -92,5 +92,7 @@ class DocumentStoreShareProvider
         error: (jqXHR) ->
           docName = metadata?.filename or 'document'
           callback "Unable to share '#{docName}'"
+    else
+      callback "Unable to unshare '#{docName}'"
```
kswenson: If there isn't already, there should be a technical debt story for moving all of the hard-coded CFM error messages into the strings file for future translatability.

PR #71, opened 2016-11-30
Fixed filename input not closing when user clicks in document [#127301439]

dougmartin:
CODAP eats the click events in the main workspace which cause the blur event not to be triggered for the filename rename input element.  This adds a global window listener to simulate the blur event.

### DIFF
--- a/src/code/views/menu-bar-view.coffee
+++ b/src/code/views/menu-bar-view.coffee
@@ -10,7 +10,7 @@ module.exports = React.createClass
   componentWillMount: ->
     # need to use direct DOM events because the event needs to be captured
     if window.addEventListener
-      window.addEventListener 'click', @checkBlur, true
+      window.addEventListener 'mousedown', @checkBlur, true
 
     @props.client._ui.listen (event) =>
       switch event.type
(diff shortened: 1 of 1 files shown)


src/code/views/menu-bar-view.coffee
```diff
@@ -8,6 +8,10 @@ module.exports = React.createClass
   displayName: 'MenuBar'
 
   componentWillMount: ->
+    # need to use direct DOM events because the event needs to be captured
+    if window.addEventListener
+      window.addEventListener 'click', @checkBlur, true
```
kswenson: This is clever. I like it. I have a couple of thoughts, however.
- Have you tested this with dragging? I would think that the click, and therefore the blur, would happen after the drag was complete. It's worth considering whether the handler should be on 'mousedown'.
- I believe the CFM menu is sometimes stickier than it should be and so a similar approach could be used there, although not necessarily part of this PR.

dougmartin:
@kswenson You are right - mousedown works much better since the user may drag a component in the workspace.  The click handler still works but the file rename input is still visible until the drag is completed and the user lifts the mouse.  I've updated the code.

PR #73, opened 2016-12-02
Fix placement of menu chevron in Firefox [#134879405]

dougmartin:


dougmartin:
@kswenson This is a big one...

dougmartin:
It now looks like this in FF

image

PR #74, opened 2016-12-06
Improve Safari Local File save [#135154489]

kswenson:
On Safari, behavior is to automatically download the file with name "Unknown". Due to limitations of Safari, this seems to be the best we can do. To accomplish this we introduce the FileSaver.js library (https://github.com/eligrey/FileSaver.js) which attempts to provide the best cross-browser support possible.

If the 'download' attribute is supported, uses the original code to download the file. If the 'download' attribute is not supported (e.g. Safari), uses the FileSaver.js library to perform the download. WebKit Nightly and the Safari Technology Preview both support the 'download' attribute, so the FileSaver.js library will only be necessary for a limited time.

In conjunction with a change to CODAP to allow import of files without extensions, CODAP will be able to open downloaded "Unknown" files even if they are not manuallly renamed.

Finally, we fix the behavior of the "Save Link As..." context menu item on supported browsers, i.e. not current versions of Safari.

### DIFF
--- a/src/code/views/local-file-tab-save-view.coffee
+++ b/src/code/views/local-file-tab-save-view.coffee
@@ -2,6 +2,7 @@
 tr = require '../utils/translate'
 CloudMetadata = (require '../providers/provider-interface').CloudMetadata
 cloudContentFactory = (require '../providers/provider-interface').cloudContentFactory
+FileSaver = require('../lib/file-saver')
 
 module.exports = React.createClass
 
@@ -19,6 +20,7 @@ module.exports = React.createClass
                   then @props.dialog.data.extension else 'json'
     state =
       filename: filename
+      supportsDownloadAttribute: document.createElement('a').download isnt undefined
       downloadFilename: @getDownloadFilename hasPropsContent, filename, extension
       extension: extension
       mimeType: if hasPropsContent and @props.dialog.data.mimeType? \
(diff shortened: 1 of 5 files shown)


src/code/views/local-file-tab-save-view.coffee
```diff
@@ -55,8 +56,9 @@ module.exports = React.createClass
 
   confirm: (e, simulateClick) ->
     if not @confirmDisabled()
-      @refs.download.href = @props.client.getDownloadUrl(@state.content, @state.includeShareInfo, @state.mimeType)
-      @refs.download.click() if simulateClick
+      blob = @props.client.getDownloadBlob(@state.content, @state.includeShareInfo, @state.mimeType)
+      FileSaver.saveAs(blob, @state.downloadFilename, true)
```
dougmartin: Do we need to move the preventDefault from the else below to above the if since we are no longer using the href transform?

kswenson:
@dougmartin Your question and the discussion in the demo meeting today made me think about this a bit differently. My new approach is to use the old technique on supported browsers (i.e. browsers that support the `download` attribute) and only use the FileSaver.js library for browsers that don't support the `download` attribute (i.e. Safari).

One other change is that we now update the download URL on `contextMenu` events as well, so "Save Link As..." works as expected on supported browsers (i.e. not Safari).

dougmartin:
Looks good 👍

PR #75, opened 2016-12-07
Validate document contents when opening a file [#134091043]

kswenson:
CFM client can pass a validator() function as part of CFM configuration. If an opened document fails validation then an error alert is shown instead of opening the file. The error message can also be specified by the CFM client, but if not specified a default message will be shown.


kswenson:
@dougmartin reminded me that the existing callback mechanism allows client-side validation to occur as long as the client uses the callback appropriately, so this change is unnecessary.

PR #76, opened 2016-12-07
Fix local file tab drop handling

kswenson:
While testing CFM PR #75 (validate document contents) I encountered the fact that files dropped in the drop zone weren't giving the expected error alert message. This turned out to be because the CODAP document drop handler was handling the 'drop' event before the React onDrop() handler. It had been noticed (but not investigated) previously that dropping a file in this dialog sometimes behaved differently than selecting a file from the file chooser and this is presumably the reason for such discrepancies. The fix is to install a DOM event handler on the element rather than using the React onDrop() mechanism.

### DIFF
--- a/src/code/views/local-file-tab-list-view.coffee
+++ b/src/code/views/local-file-tab-list-view.coffee
@@ -6,6 +6,16 @@ module.exports = React.createClass
 
   displayName: 'LocalFileListTab'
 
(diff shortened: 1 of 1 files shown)


dougmartin:
Looks good 👍 Sorry - I forgot about this one yesterday

PR #77, opened 2016-12-08
Fixed optional callback during re-authorization [#128473793]

dougmartin:


dougmartin:
Yes - this was meant for all the CFM fixes but only one made it in :)

PR #78, opened 2016-12-14
LaraProvider should disable autosave with read-only access key (again)

kswenson:
@scytacki did all the hard work:

> When viewing one of these read-only links, when the user makes a change (or if the data interactive triggers a change). Then the error message:
> "Unable to save undefined: [error.missingParam]" is shown.
> 
> In this document the data interactive causes CODAP to try to autosave right away.  This issue was fixed previously in this PR: https://github.com/concord-consortium/cloud-file-manager/pull/58.

> I think it is caused by the change to 'resave' [here](https://github.com/concord-consortium/cloud-file-manager/blob/master/src/code/client.coffee#L339)
> Which wasn't added [here](https://github.com/concord-consortium/cloud-file-manager/blob/master/src/code/providers/lara-provider.coffee#L81)


dougmartin:
Looks good 👍

PR #80, opened 2016-12-22
Fix double-download on Local File Save

kswenson:
Use browser events for the download button rather than React events for predictable behavior [#136287689]

The onClick handler for the download button was getting triggered multiple times and preventDefault seemed not to work as expected. The resulting behavior varied between browsers and sometimes between dev and non-dev builds. After spending too much time trying to fix things in the React SyntheticEvent system, switching to browser event handlers eliminated the oddities.


dougmartin:
Looks good 👍

PR #81, opened 2016-12-22
Eliminate react warnings

kswenson:
Eliminate `setState called on unmounted component` React warnings
- use synchronized instance variable and state variable in GoogleDriveProvider and AuthorizeMixin
- added _isMounted instance variable to handle cases in which component is unmounted when asynchronous call returns

I didn't make the corresponding changes in the DocumentStoreProvider, since that's going away imminently.

Note: It's possible that some of the Google Drive authorization-related oddities that have been reported could be affected by authorization state-handling, so some testing of those issues would make sense.

### DIFF
--- a/src/code/providers/google-drive-provider.coffee
+++ b/src/code/providers/google-drive-provider.coffee
@@ -12,8 +12,7 @@ GoogleDriveAuthorizationDialog = React.createFactory React.createClass
   displayName: 'GoogleDriveAuthorizationDialog'
 
   getInitialState: ->
-    @_hasLoadedGAPI = false
-    loadedGAPI: false
+    loadedGAPI: window._LoadedGAPIClients
 
   # See comments in AuthorizeMixin for detailed description of the issues here.
   # The short version is that we need to maintain synchronized instance variable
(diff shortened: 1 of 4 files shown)


src/code/providers/google-drive-provider.coffee
```diff
@@ -195,8 +211,10 @@ class GoogleDriveProvider extends ProviderInterface
   _loadGAPI: ->
     if not window._LoadingGAPI
       window._LoadingGAPI = true
-      window._GAPIOnLoad = ->
-        @window._LoadedGAPI = true
+      window._GAPIOnLoad = =>
+        window._LoadedGAPI = true
+        # preload clients to avoid user delay later
+        @_loadedGAPI ->
```
dougmartin: You set @_loadedGAPI to false above but `@_loadedGAPI ->` will transpile to a function call like this `this._loadedGAPI(function() {});`

src/code/providers/google-drive-provider.coffee
```diff
@@ -12,11 +12,27 @@ GoogleDriveAuthorizationDialog = React.createFactory React.createClass
   displayName: 'GoogleDriveAuthorizationDialog'
 
   getInitialState: ->
+    @_hasLoadedGAPI = false
```
dougmartin: I think you could replace this with the existing ` window._LoadedGAPI` variable

kswenson: I was following the pattern I used in `AuthorizeMixin`, but what you suggest would be simpler so I'll check it out.

src/code/providers/google-drive-provider.coffee
kswenson: The intent was to call the `_loadedGAPI()` method with an empty callback. No instance variables were set on the GoogleDriveProvider itself, and the provider isn't in a position to easily reach across to the AuthorizationDialog. In any case, per your suggestion above the AuthorizationDialog is using the global variable now and so there shouldn't be anything that needs to be reset in the callback.

src/code/providers/google-drive-provider.coffee
kswenson: It's actually the clients that we're interested in, so it's `window._LoadedGAPIClients` that we care about, but using that simplifies the code. I've added another commit which makes the change.

src/code/providers/google-drive-provider.coffee
dougmartin: Sorry I missed the underscore - I thought you were referencing the new variable you created on line 15.

PR #82, opened 2017-02-16
Added support to export data via postmessage instead of the Save As file dialog [#138395609]

dougmartin:
This support is enabled with the existence of the new saveSecondaryFileViaPostMessage query parameter.

### DIFF
--- a/src/code/client.coffee
+++ b/src/code/client.coffee
@@ -548,15 +548,23 @@ class CloudFileManagerClient
       callback? 'No initial opened version was found for the currently active file'
 
   saveSecondaryFileAsDialog: (stringContent, extension, mimeType, callback) ->
-    data = { content: stringContent, extension, mimeType }
-    @_ui.saveSecondaryFileAsDialog data, (metadata) =>
-      # replace defaults
-      if extension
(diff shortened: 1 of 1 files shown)


src/code/client.coffee
```diff
@@ -548,15 +548,23 @@ class CloudFileManagerClient
...
-    data = { content: stringContent, extension, mimeType }
-    @_ui.saveSecondaryFileAsDialog data, (metadata) =>
-      # replace defaults
-      if extension
-        metadata.filename = CloudMetadata.newExtension metadata.filename, extension
-      if mimeType
-        metadata.mimeType = mimeType
-
-      @saveSecondaryFile stringContent, metadata, callback
+    if window.location.search.indexOf("saveSecondaryFileViaPostMessage") isnt -1
```
kswenson: Is there a reason to prefer URL query parameter here rather than using a property in the CFM config?

dougmartin: The "sharinator" needs this functionality when it iframes CODAP so an url parameter is the best option.  The other options are to use either a unique CODAP build or a unique CFM build and pass the baseurl to CODAP -- both of which means keeping the dependencies up to date.

kswenson: 👍

PR #83, opened 2017-02-24
Hamburger menu too close to doc name for touch

kswenson:
Fixes [#137658223] (from Bill Finzer, who originally made the change locally within CODAP)
Increase padding-right for the menu and decrease padding-left for the file menu


kswenson:
Reviewed in real time with Jonathan

PR #84, opened 2017-04-19
Added postmessage listener

dougmartin:
Added a postmessage listener that currently consists of two message handlers:

1. getCommands, which returns an array of commands that the postmessage api handles
2. autosave, which autosaves the document if it is able to


dougmartin:
Related docstore PR: https://github.com/concord-consortium/document-store/pull/23

PR #90, opened 2017-09-07
Support POEditor; add Turkish & Hebrew translations; auto-detect language

kswenson:
- Changes to master English strings file can be pushed to POEditor with `npm run poeditor:push`. (POEditor API key must be specified as command-line argument or set as environment variable. It is too sensitive to be committed to Github.)
- Currently, master English strings file cannot contain comments. If comments are desired, an additional script step can be used, such as the one currently being used by Building Models.
- Updated translations can be pulled from POEditor with `npm run poeditor:pull`. (POEditor API key must be specified as command-line argument or set as environment variable. It is too sensitive to be committed to Github.) Note that the pull command also copies the master English strings file into its JavaScript form in the project, so it should be used to update the English as well.
- The CFM is now configured to support multiple languages and to auto-detect the correct language, so only one generic copy of the CFM is required rather than a separate copy for each language.
- Testable (in CODAP) at http://codap.concord.org/~kswenson/translate/static/dg/tr/cert/index.html (Turkish) and http://codap.concord.org/~kswenson/translate/static/dg/he/cert/index.html (Hebrew).

@dougmartin Don't merge until we've gotten more feedback about the overall approach.


bin/poeditor-pull.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash                                                                                                                                                                                     
+CURL='/usr/bin/curl'
+POEDITOR_EXPORT_URL="https://api.poeditor.com/v2/projects/export"
+API_TOKEN="$POEDITOR_API_TOKEN" # may be set as environment variable
```
dougmartin: I think you should add something like this to ensure the API_TOKEN has a value

`[[ -z "$API_TOKEN" ]] && { echo "No API_TOKEN available!" ; exit 1; }`

bin/poeditor-pull.sh
```diff
@@ -0,0 +1,44 @@
...
+    ;;
+esac
+shift # past argument or value
+done
+
+CURLARGS="-X POST -d id=$PROJECT_ID -d language=$LANGUAGE -d type=key_value_json -d api_token=$API_TOKEN"
+# echo "CURLARGS='$CURLARGS'"
+
+# retrieve the initial API response
+RESPONSE=$($CURL $CURLARGS $POEDITOR_EXPORT_URL)
```
dougmartin: You can add the same check as above to each of the next 3 lines

`[[ -z "$VARIABLE_NAME" ]] && { echo "VARIABLE_NAME is empty" ; exit 1; }`

PR #91, opened 2017-09-27
Optionally pass LARA information back to client for logging

kswenson:
Optionally pass LARA information back to client for logging [#151493445]
passes the following information (if known)
- run state URL
- document store document ID
- document store document URL
- collaborator URLs

Note that this relies on the client providing a `logLaraData` callback, which will be added to CODAP in [CODAP PR #202](https://github.com/concord-consortium/codap/pull/202).

### DIFF
--- a/src/code/providers/lara-provider.coffee
+++ b/src/code/providers/lara-provider.coffee
@@ -62,6 +62,9 @@ class LaraProvider extends ProviderInterface
     else
       false
 
+  logLaraData: (laraData) ->
+    @options.logLaraData laraData if @options.logLaraData
+
   # don't show in provider open/save dialogs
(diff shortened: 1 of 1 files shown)


src/code/providers/lara-provider.coffee
```diff
@@ -62,6 +62,10 @@ class LaraProvider extends ProviderInterface
     else
       false
 
+  logLaraData: (laraData) ->
+    logLaraData.collaboratorUrls = @collaboratorUrls if @collaboratorUrls?.length
```
scytacki: I'm probably not understanding this right. But should this be setting the collaboratorUrls on the laraData object instead of the logLaraData object?

kswenson:
Good catch. Fixed typo.

PR #92, opened 2017-10-30
"Headless" CFM

kswenson:
Make it easier for third parties (e.g. Zoom In Science) to work with CFM:
- eliminate hard-coded provider dependencies
- make it possible to avoid user-facing dialogs with appropriate providers

Only instantiate `DocumentStoreShareProvider` when using the `DocumentStoreProvider` [#152333125]

Add `autoProvider` architecture
- providers can support a capability without user intervention (i.e. without dialogs)
- `saveContent()` and `saveSecondaryFileAsDialog()` methods check for `autoProvider` [#152332814]
- move `saveSecondaryFileViaPostMessage` capability into new PostMessageProvider

### DIFF
--- a/src/code/client.coffee
+++ b/src/code/client.coffee
@@ -1,6 +1,7 @@
 tr = require './utils/translate'
 isString = require './utils/is-string'
 base64Array = require 'base64-js' # https://github.com/beatgammit/base64-js
+getQueryParam = require './utils/get-query-param'
 
 CloudFileManagerUI = (require './ui').CloudFileManagerUI
 
@@ -11,6 +12,7 @@ LaraProvider = require './providers/lara-provider'
 DocumentStoreProvider = require './providers/document-store-provider'
 DocumentStoreShareProvider = require './providers/document-store-share-provider'
 LocalFileProvider = require './providers/local-file-provider'
+PostMessageProvider = require './providers/post-message-provider'
 URLProvider = require './providers/url-provider'
 
 cloudContentFactory = (require './providers/provider-interface').cloudContentFactory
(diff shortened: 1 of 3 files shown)


src/code/providers/post-message-provider.coffee
```diff
@@ -0,0 +1,31 @@
...
+        export: if getQueryParam "saveSecondaryFileViaPostMessage" then 'auto' else false
+        load: false
+        list: false
+        remove: false
+        rename: false
+        close: false
+
+  canOpenSaved: -> false
+
+  save: (content, metadata, callback) ->
```
dougmartin: It might be better to rename this `export` since that is the capability it is providing.

src/code/client.coffee
```diff
@@ -571,15 +591,11 @@ class CloudFileManagerClient
   # Saves a file to backend, but does not update current metadata.
   # Used e.g. when exporting .csv files from CODAP
   saveSecondaryFile: (stringContent, metadata, callback = null) ->
-    if metadata?.provider?.can 'save', metadata
-      @_setState
-        saving: metadata
+    if metadata?.provider?.can 'export', metadata
       metadata.provider.save stringContent, metadata, (err, statusCode) =>
```
dougmartin: See comment in `post-message-provider.coffee` about renaming this to `export`

src/code/client.coffee
```diff
@@ -39,7 +41,16 @@ class CloudFileManagerClient
 
     # filter for available providers
     allProviders = {}
-    for Provider in [ReadOnlyProvider, LocalStorageProvider, GoogleDriveProvider, LaraProvider, DocumentStoreProvider, LocalFileProvider]
+    providerList = [
```
dougmartin: I think you should add a note here about the order of providers being important when it searches for an auto provider in `autoprovider()` as it returns the first one that marks the capability as `auto`.  It might help in the future if other providers are added that mark the same capabilities as `auto`.

kswenson: Done. It practice I don't think it will matter much as the interest in auto-providers mostly comes from clients who are configuring the CFM to only have a single provider, but the comment makes sense in any case.

src/code/providers/post-message-provider.coffee
kswenson: I thought a little bit about that. I wanted to preserve the pre-existing notion that export/save are separate capabilities that can share an implementation, but as your comment suggests that implicit distinction was not entirely clear. With the latest commit I've made it explicit by adding the `saveAsExport()` method (I avoided `export` as it's a reserved word in JavaScript) along with a default implementation that defers to the `save()` method, so existing clients that support both capabilities but only implement `save()` will continue to work.

kswenson:
@dougmartin I've responded to your code review comments. Care to re-review? Thx

dougmartin:
Looks good 👍

PR #95, opened 2017-11-01
Support logging callback

kswenson:
Support `log` callback for logging through client application [#152333085]
- add `save` logging to `DocumentStoreProvider` and `LaraProvider` for debugging purposes

Eliminate trailing slashes from `@docStoreUrl`
- having recently encountered situations in which redundant slashes caused problems for some servers, we proactively remove the potential problem.

@scytacki Does this look like the right information to log to help with debugging the document corruption issue?

### DIFF
--- a/src/code/client.coffee
+++ b/src/code/client.coffee
@@ -175,6 +175,10 @@ class CloudFileManagerClient
     if listener
       @_listeners.push listener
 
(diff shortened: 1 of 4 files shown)


scytacki:
Looks great to me. 👍

PR #97, opened 2017-12-15
Reduce `drive` permission to `drive.readonly`

sfentress:
`drive.readonly` gives us permission to read all files, for the file
dialog, and we already have write permissions to the files we create
through the `drive.file` permission.

### DIFF
--- a/src/code/providers/google-drive-provider.coffee
+++ b/src/code/providers/google-drive-provider.coffee
@@ -97,7 +97,7 @@ class GoogleDriveProvider extends ProviderInterface
       args =
         client_id: @clientId
         scope: [
-          'https://www.googleapis.com/auth/drive'
+          'https://www.googleapis.com/auth/drive.readonly'
           'https://www.googleapis.com/auth/drive.install'
           'https://www.googleapis.com/auth/drive.file'
           'https://www.googleapis.com/auth/userinfo.profile'


jsandoe:
@sfentress I noticed that this CFM change from 12/15 had not been incorporated in CODAP, so I pulled it. When I try to open Google Drive with this change in place I get the following:

pasted image at 2018_01_16 03_15 pm

I think Google Drive has detected a change, which has caused them to treat the app as untrusted.
I don't know whether there is something we can do on the admin page to avoid this. Also, was this change motivated by a recommendation from Google or some other source? If so, I would be curious whether they mention this consequence. 

I don't have access, so far as I know, to the Google Drive admin for the App.  If its easy to convey this to me, I would be happy to attempt to follow up, myself.

sfentress:
Hi @jsandoe,

This warning message has actually been visible for all new users of CODAP for a while now, the first time they go to authorize the app with their Drive account. I noticed it when I deauthorized my account last month, and then tried to save a new file to Drive.

Did you de-authorize your account before seeing this?

I don't see that warning when I go to save a document to Drive from CODAP right now, presumably because I hit "OK" last month when I tried it, so I don't think that the permissions change triggered that for everybody.

As for when Google will reauthorize us, pulling in this PR was part of that work, as Google specifically requested that we change it. They also requested that we remove any http cross-origin requests, which I think @kswenson was working on? When that is done, @jengoree can reply to their email, and hopefully they will reauthorize us.

PR #101, opened 2018-03-23
Update LARA share dialog to use autolaunch page and fullscreen scaling

pjanik:
Changes:
- switch default page to "autolaunch"
- add "autolaunch" / "launch" page checkbox
- add "fullscreen scaling" checkbox (it passes "scaling" URL param to doc-server, which modifies CODAP iframe to scale down and include fullscreen button)
- display inputs bases on page tyle (Launch Button Text for "launch" page, scaling for "autolaunch")
- cleanup and simplify some related code (e.g. `lara` state was redundant, it can be generated using other properties; simplify some event handlers, remove refs)

Screenshots:
screen shot 2018-03-23 at 16 45 11
screen shot 2018-03-23 at 16 45 18

@dougmartin or @kswenson, could you take a look? @dougmartin seems to be the main contributor, but @kswenson works on IS2 project. 😉

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ node_modules/
 
 # OS or Editor folders
 .DS_Store
+.idea
--- a/src/code/views/share-dialog-view.coffee
+++ b/src/code/views/share-dialog-view.coffee
@@ -42,11 +42,10 @@ module.exports = React.createClass
   getInitialState: ->
     link: @getShareLink()
     embed: @getEmbed()
-    lara: @getLara
-      codapServerUrl: "https://codap.concord.org/releases/latest/"
-      launchButtonText: "Launch"
+    pageType: "autolaunch"
     codapServerUrl: "https://codap.concord.org/releases/latest/"
     launchButtonText: "Launch"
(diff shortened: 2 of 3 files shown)


ddamelin:
I didn't realize you would be in this code. If I had known I would have requested you also add the option to have the app=is URL parameter as a checkbox as well. This causes the parent visibility toggles to appear in graphs be default. We typically use this parameter for sensor-based experiments, simulations, and SageModeler documents. It can be added in to the CODAP server URL by hand, but it would be easier for non-technical people to just check a box to add it there.

jsandoe:
This PR missed the CODAP build today. I hope that it is not urgent. I generally check that the CFM has no changes on master when I do a CODAP build. Thus, I didn't see this.

ddamelin:
It's not urgent but would be good to get into the next build if possible.

ddamelin:
@pjanik The name would be "display data visibility toggles on graphs" and it should be off by default.

jsandoe:
Am merging this PR so it can make the next CODAP build.

PR #102, opened 2018-04-17
Fix copy link in LARA share dialog

pjanik:
It got broken as I removed `lara` from the state. I'm betting `embed` and `link` could also be removed in the future if they can be calculated at any time (it seems so). Keeping them in the state is redundant.

[#156243160]


pjanik:
I've added one more commit that fixes handling of the graph visibility toggles param (`app=is`). It was added to document-store URL, but it should have been part of the CODAP server URL.

scytacki:
@jsandoe, @kswenson can one of you merge this?

PR #105, opened 2018-07-11
[#153377116] Embedded mode CODAP should not set page title.

jsandoe:
We remove the responsibility to set the page title from the CFM and
restore it to CODAP.

The CFM had been setting title element for the containing page whenever the name of the current document changed. This causes problems for the shared form of Embedded Mode and would cause problems in other modes. Move the responsibility of managing the title element to the hosting app (CODAP).

### DIFF
--- a/src/code/client.coffee
+++ b/src/code/client.coffee
@@ -105,9 +105,6 @@ class CloudFileManagerClient
       shareProvider: shareProvider
 
     @appOptions.ui or= {}
-    @appOptions.ui.windowTitleSuffix or= document.title
-    @appOptions.ui.windowTitleSeparator or= ' - '
-    @_setWindowTitle()
 
     @_ui.init @appOptions.ui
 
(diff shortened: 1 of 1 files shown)


jsandoe:
I agree there is, for now, a concern about Sage. I have been thinking of creating a permanent CODAP fork of CFM for CODAP-only fixes. Sage could then pull particular commits into the master codeline as it sees fit.

ddamelin:
I'm not sure I understand the implications for SageModeler in standalone mode. In that case it is basically the same as a plugin, but just taking up the entire canvas. It's relationship to CFM/CODAP is the same AFAIK.

kswenson:
@ddamelin By standalone Sage/CFM in this case we mean without CODAP, not CODAP's "standalone" mode. (We need better terminology for these things.) The change in this PR moves responsibility for setting the window title from CFM to the client app. CODAP is presumably setting the window title as it deems appropriate. In the standalone Sage/CFM case (without CODAP), Sage may need to set the window title itself if it cares to, since CFM is no longer doing so.

jsandoe:
Merged to CODAP branch to avoid breaking Sage when Sage is used outside CODAP.

PR #106, opened 2018-07-20
Fix inconsistent behavior of CFM menu

kswenson:
Fixes inconsistent behavior of CFM menu [#158700996]
- sometimes clicking on an item with a submenu would close the menu rather than opening the submenu
- sometimes clicking on the "anchor" would fail to open or close the menu appropriately
- the original bug was listed as being touch related but in many cases it was more timing-related
- the fix here is to eliminate the timer mechanism for closing the menu in favor of a class-based one

### DIFF
--- a/src/code/views/dropdown-view.coffee
+++ b/src/code/views/dropdown-view.coffee
@@ -65,6 +65,8 @@ DropDown = React.createClass
       window.removeEventListener 'touchstart', @checkClose, true
 
   checkClose: (evt) ->
+    # no need to walk the DOM if the menu isn't open
+    return if not @state.showingMenu
     # if the click is on the menu, let the menu handle it
     elt = evt.target
     while elt?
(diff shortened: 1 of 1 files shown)


src/code/views/dropdown-view.coffee
```diff
@@ -54,27 +56,22 @@ DropDown = React.createClass
...
-
-  unblur: ->
-    if @state.timeout
-      clearTimeout(@state.timeout)
-    @setState {timeout: null}
+      window.removeEventListener 'mousedown', @checkClose, true
+      window.removeEventListener 'touchstart', @checkClose, true
+
+  checkClose: (evt) ->
+    # if the click is on the menu, let the menu handle it
```
dougmartin: Since this is called for every click in the window I wonder if we should short circuit and exit if `@state.showingMenu` is `false` - that will avoid the walking the DOM tree for each click.

kswenson: Excellent point! Done.

PR #107, opened 2018-07-20
Update dependencies

kswenson:
- recent CODAP builds have used newer versions than those specified in package.json


kswenson:
I stopped at 2.3.2 because the 2.4.x line seemed to mostly be changing bundle settings and generating a lot of churn that hadn't quieted down yet.

jsandoe:
Thats what I had figured.

PR #113, opened 2018-11-20
Add language menu support

pjanik:
This PR makes Dropdown component more flexible and reuses it to render the language menu.


pjanik:
Ops, thanks. It should be fixed now.

PR #115, opened 2018-11-28
Fixed event-stream dep vuln

dougmartin:
Updated event-stream dep and replaced gulp-inject-string with inlined
code as it depended on an old event-stream.


package.json
```diff
@@ -29,14 +29,13 @@
     "coffee-script": "^1.10.0",
     "coffeeify": "^2.0.1",
     "del": "^2.1.0",
-    "event-stream": "^3.3.0",
+    "event-stream": "^3.3.4",
```
kswenson: To reduce the likelihood of someone inadvertently bumping to a newer vulnerable version, I would make this `3.3.4` rather than `^3.3.4`.

PR #127, opened 2019-04-10
Added scropes option to google-drive-provider

dougmartin:
This allows sage-modeler-site to specify a resticted list of scopes.

### DIFF
--- a/src/code/providers/google-drive-provider.coffee
+++ b/src/code/providers/google-drive-provider.coffee
@@ -69,6 +69,12 @@ class GoogleDriveProvider extends ProviderInterface
     @clientId = @options.clientId
     if not @clientId
       throw new Error (tr "~GOOGLE_DRIVE.ERROR_MISSING_CLIENTID")
+    @scopes = @options.scopes or [
+      'https://www.googleapis.com/auth/drive'
+      'https://www.googleapis.com/auth/drive.install'
+      'https://www.googleapis.com/auth/drive.file'
+      'https://www.googleapis.com/auth/userinfo.profile'
+    ]
     @mimeType = @options.mimeType or "text/plain"
     @readableMimetypes = @options.readableMimetypes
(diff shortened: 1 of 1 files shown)


src/code/providers/google-drive-provider.coffee
```diff
@@ -69,6 +69,12 @@ class GoogleDriveProvider extends ProviderInterface
     @clientId = @options.clientId
     if not @clientId
       throw new Error (tr "~GOOGLE_DRIVE.ERROR_MISSING_CLIENTID")
+    @scopes = @options.scopes or [
+      'https://www.googleapis.com/auth/drive'
```
kswenson: Late to the party here, but I was just looking up the meaning of these scopes out of curiosity. Is this a reasonable default for the CFM? Does CODAP/CFM need this scope? Given that the Google docs describe it as
>Full, permissive scope to access all of a user's files, excluding the Application Data folder. Request this scope only when it is strictly necessary.

If Sage/CFM can get by with `drive.metadata.readonly` rather than `drive`, could CODAP/CFM do so as well? Independent of the answer to that question, it seems like CFM should default to the minimum set of scopes and then applications that require more can request it rather than having CFM ask for more and have clients that require less have to ask for it. If any of this is worthy of further consideration, perhaps a PT story should be written on the CODAP side. cf. @jsandoe

dougmartin: CODAP uses the full scopes so it could read .codap and non .codap files and was granted that access by Google.  I had to make this change quickly as Dan is demoing this on Saturday to a large number of people so this was the safest way to do it.

kswenson: Completely agree that your change needed to be made the way you did so as not to break CODAP. Now that the issue has been raised, however, we can reasonably ask whether there's a better way for CODAP/CFM to handle these properties moving forward.

PR #136, opened 2019-11-19
Removed Safari specific note about naming files when saving locally [#169586400]

dougmartin:


dougmartin:
It is no longer required in Safari and we don't know what version of Safari fixed it so Dan wanted it removed all together.

PR #143, opened 2019-11-26
Fix extension not being added when filename as dot in it [#168903516]

dougmartin:
Fixes two bugs:

1. The check for the existing extension was comparing the substring at the end of the name of length extensionLength -1 to the concatenaton of a dot and the extension instead of the substring of extensionLength + 1 (the +1 to add the dot being compared).

2. When the extension was added any existing extension was being dropped with no check to see if the extensions matched.

This change also tries to unpack the dense update code with comments and a conditional path.


dougmartin:
@dstrawberrygirl I forgot to assign this to you last week.

PR #149, opened 2020-03-17
Merge master branch changes to CODAP branch

kswenson:
Preparation for upcoming React update (#148) which will also need to be merged to the CODAP branch.
- @dougmartin Does it look like I retained the `master` branch changes?
- @jsandoe Does it look like I retained the `CODAP` specializations? Is there anything here that shouldn't be merged to the CODAP branch?

Ideally, we would turn the remaining `CODAP` branch specializations into configurations so that the CODAP build could be built off the `master` branch. Some efforts have been made along these lines (cf. #110 and [CODAP #241](https://github.com/concord-consortium/codap/pull/241)) but things have no doubt diverged further since then.

### DIFF
--- a/src/code/client.coffee
+++ b/src/code/client.coffee
@@ -136,6 +136,9 @@ class CloudFileManagerClient
     @newFileOpensInNewTab = if @appOptions.ui?.hasOwnProperty('newFileOpensInNewTab') then @appOptions.ui.newFileOpensInNewTab else true
     @newFileAddsNewToQuery = @appOptions.ui?.newFileAddsNewToQuery
 
+    if @appOptions.ui?.confirmCloseIfDirty
+      @_setupConfirmOnClose()
+
     @_startPostMessageListener()
 
   setProviderOptions: (name, newOptions) ->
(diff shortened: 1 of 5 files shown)


src/code/client.coffee
```diff
@@ -136,6 +136,9 @@ class CloudFileManagerClient
     @newFileOpensInNewTab = if @appOptions.ui?.hasOwnProperty('newFileOpensInNewTab') then @appOptions.ui.newFileOpensInNewTab else true
     @newFileAddsNewToQuery = @appOptions.ui?.newFileAddsNewToQuery
 
+    if @appOptions.ui?.confirmCloseIfDirty
+      @_setupConfirmOnClose()
+
```
jsandoe: CODAP already confirms close on dirty documents. It must use a different mechanism. I realize this feature is enabled with a flag so there isn't a danger of conflict, but should there be two mechanisms to accomplish the same thing?

kswenson: Looking at the CODAP implementation, it appears that the CODAP requirements may be more elaborate than the CFM implementation, so there may be a reason to have two implementations.
```
if (DG.currDocumentController().get('hasUnsavedChanges') &&
    (DG.get('embeddedMode') === 'no') &&
    (DG.get('componentMode') === 'no')) {
  return 'DG.AppController.beforeUnload.confirmationMessage'.loc();
}
```

PR #157, opened 2020-04-22
172226702 switch to javascript

knowuh:
* Automatic coffeescript → javascript convsion using bulk `bulk-decaffeinate convert`
* Add some lint rules to help find regressions & issues
* Manual fixes, mostly from coffeescript constructors using this before calling super.
* Remove mixin
* Address easily addressable lint errors

…


knowuh:
@dougmartin this is mostly the automated conversion, with just a little fixup.

I didn't go very far down the JSX path, I only converted the top level component.

I will do more fixes after this work has been reviewed, so the changes are more clear, and easier to review.

src/code/views/menu-bar-view.js
```diff
@@ -0,0 +1,219 @@
...
+  displayName: 'MenuBar',
+
+  componentDidMount() {
+    // need to use direct DOM events because the event needs to be captured
+    if (window.addEventListener) {
+      window.addEventListener('mousedown', this.checkBlur, true)
+      window.addEventListener('touchstart', this.checkBlur, true)
+    }
+
+    return this.props.client._ui.listen(event => {
```
dougmartin, line 31: Just a note for future TypeScript work.  All the _ variables were meant to be private but as you can see the `_ui` variable leaks out in various places.  It might make sense to refactor who we use it in other files and make those methods public.

tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
```
dougmartin, line 1: For faster builds you may want to enable the `incremental` config option (https://www.typescriptlang.org/v2/en/tsconfig#incremental).

PR #158, opened 2020-04-28
Add Typing for the provider-interface

knowuh:
[#172557566]
https://www.pivotaltracker.com/story/show/172557566

[#172226702]
https://www.pivotaltracker.com/story/show/172226702

First pass on cleaning up the provider-interface

* Tried to extract insert some helpful types
* Attended to a few decaffeinate Todos


knowuh:
You can ignore `provider-interface-original.js` I want to keep it around to run focused tests against.

@dougmartin  Please have a look at a comment / question I have on line 109 in `provider-interface.ts` 

I am planning at least one or two smaller incremental PRs that tighten up this interface and add unit tests, but I am trying to keep the PR sizes down for sanity.

src/code/providers/provider-interface.ts
```diff
@@ -0,0 +1,550 @@
...
+type providerType = string
+type parentType = string // Based on the code, seems like this might path?
+type providerDataType = any  // Not sure what this is
+type documentId = string
+type key = string
+type mimeType = string
+
+
+// TODO: What does a save callback signature really looklike?
+type ISaveCallback = (arg: any) => any
```
dougmartin, line 26: I'd suggest type aliasing the `any` s that you want to eventually tighten up to a concrete type so that you can search for them later as some of the `any` types should stay `any`.  Something like:

```
type IAnyForNow = any
type ISaveCallback = (arg: IAnyForNow) => IAnyForNow
```

dougmartin: If you put that `IAnyForNow` in an importable module then you can use VS Code's "find all uses" command.

dougmartin: One other thing - in other projects we have been reserving the `I` prefix for interfaces and not using it on types.

src/code/providers/provider-interface.ts
```diff
@@ -0,0 +1,550 @@
...
+  ReadableExtensions?: string[]
+}
+interface ISharedMetadata {
+  _permissions?: any,
+  shareEditKey?: string,
+  sharedDocumentId?: string,
+  accessKeys?: string | string[]
+}
+
+type CloudFileContentType = any
```
dougmartin, line 76: I think it is safe to make this a `string` instead of `any`.

src/code/providers/provider-interface.ts
```diff
@@ -0,0 +1,550 @@
...
+  sharedContentId: documentId
+  sharedContentSecretKey: key
+  mimeType: mimeType
+  filename: string|null
+
+  static Folder = 'folder'
+  static File = 'file'
+  static Label = 'label'
+
+  // TODO IMPORTANT: These are defined as class variables,
```
dougmartin, line 109: Yeah, I'm not sure.  I think it will become obvious once you start typing the providers that use this.

src/code/providers/provider-interface.ts
```diff
@@ -0,0 +1,550 @@
...
+    }
+    return _path
+  }
+
+  rename(newName: string) {
+    this.name = newName
+    return this._updateFilename()
+  }
+
+  _updateFilename() {
```
dougmartin, line 182: I would try to make all the underscore methods and variables protected (in this case as it is a base class but private in terminal classes) and see if they are leaking out.

src/code/providers/provider-interface.ts
```diff
@@ -0,0 +1,550 @@
...
+  content: any
+  // If we are 'wrapped' data & meta-data are both top-level
+  [key: string]: any
+}
+class CloudContent implements ICloudContentInterface{
+  static wrapFileContent: boolean = true
+
+  content : any
+  contentFormat: any
+  [key: string]: any
```
dougmartin, line 307: I didn't know you could do this - interesting.

src/code/providers/provider-interface.ts
```diff
@@ -0,0 +1,550 @@
...
+  constructor(_content:any, _contentFormat:any) {
```
dougmartin, line 308: I think you can rename `_content` to `content` so it doesn't look like a private variable, same with `_contentFormat`.

src/code/providers/provider-interface.ts
```diff
@@ -0,0 +1,550 @@
...
+  resave='resave',
+  load='load',
+  list='list',
+  remove='remove',
+  rename='rename',
+  close='close'
+}
+
+type IProviderCapabilities = {
+  [c in ECapabilities]?: boolean | 'auto'
```
dougmartin, line 403: Also very interesting.

PR #159, opened 2020-04-28
Test save restore all providers

eireland:
Added tests for save and restore for local storage and local file, but not Google Drive
Added tests for sharing and unsharing
Added test for import, and read-only/example docs


knowuh:
Thanks @eireland  this looks great.  Only comments I had were WRT comments. I am going to merge it as it is.

PR #161, opened 2020-04-29
Add jest testing and linting before travis deploy

knowuh:
Adding testing framework before I continue with typing and cleanup.

@eireland I added jest here, but there is another fix in this PR that might help us with some of the cypress errors we saw.

### DIFF
--- a/src/code/client.js
+++ b/src/code/client.js
@@ -34,7 +34,8 @@ import { reportError } from "./utils/report-error"
 
 
 let CLOUDFILEMANAGER_EVENT_ID =0
-const CLOUDFILEMANAGER_EVENTS   ={}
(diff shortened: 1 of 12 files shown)


src/code/client.js
```diff
@@ -194,21 +189,19 @@ class CloudFileManagerClient {
...
-      for (let provider of Array.from(this.state.availableProviders)) {
-        if (provider.name === name) {
-          if (provider.options == null) { provider.options = {} }
-          for (let key in newOptions) {
-            provider.options[key] = newOptions[key]
-          }
-          break
-        } else {
-          result.push(undefined)
+    const result = []
```
dougmartin, line 192: I think you can just get rid of the result return.  That was just there because everything is a expression in CoffeeScript.

PR #163, opened 2020-05-06
172226702 final cleanup from decaffinate

knowuh:
This should wrap up the Coffeescript conversion.

@dougmartin if you could have a quick peek, most of this is deleted `*.coffee` files but there is a little manual cleanup.  

It should be noted that there are still some TODOs in the source code which I hope we can get to in ad hoc fashion as we continue working.  I want to close the PT story: [Migrate CFM from coffeescript to javascript](https://www.pivotaltracker.com/story/show/172226702)


src/code/client.js
```diff
@@ -909,10 +907,13 @@ class CloudFileManagerClient {
...
 
   shouldAutoSave() {
-    return this.state.dirty &&
-      !(this.state.metadata != null ? this.state.metadata.autoSaveDisabled : undefined) &&
-      !this.isSaveInProgress() &&
-      __guard__(this.state.metadata != null ? this.state.metadata.provider : undefined, x => x.can('resave', this.state.metadata))
+    const { metadata } = this.state
+    return (
+      this.state.dirty
+      &&!(metadata != null ? metadata.autoSaveDisabled : undefined)
```
dougmartin, line 913: could use `!metadata?.autoSaveDisabled` here

src/code/client.js
```diff
@@ -1132,14 +1134,25 @@ class CloudFileManagerClient {
...
+            : tr("~MENUBAR.UNTITLED_DOCUMENT")
+          document.title = `${displayName}${windowTitleSeparator}${windowTitleSuffix}`
+      }
     }
   }
 
   _getHashParams(metadata) {
-    let openSavedParams
-    if (__guard__(metadata != null ? metadata.provider : undefined, x => x.canOpenSaved()) && ((openSavedParams = __guard__(metadata != null ? metadata.provider : undefined, x1 => x1.getOpenSavedParams(metadata))) != null)) {
+    let openSavedParams = metadata?.provider?.getOpenSavedParams(metadata) || null
```
dougmartin, line 1153: I think you can get rid of the `|| null` as `metadata?.provider?.getOpenSavedParams(metadata)` will return either the params or undefined.

src/code/client.js
```diff
@@ -1170,9 +1183,11 @@ class CloudFileManagerClient {
             const callbackArgs = JSON.stringify(Array.prototype.slice.call(arguments))
             return reply('cfm::event:reply', {eventId: data.eventId, callbackArgs})
         })
-        case 'cfm::event:reply':
-          var event = CLOUDFILEMANAGER_EVENTS[data.eventId]
-          return __guard__(event != null ? event.callback : undefined, x => x.apply(this, JSON.parse(data.callbackArgs)))
+        case 'cfm::event:reply': {
+          const event = CLOUDFILEMANAGER_EVENTS[data.eventId]
+          const callbackData = JSON.parse(data.callbackArgs)
```
dougmartin, line 1188: If `data.callbackArgs` is `undefined` this will throw a parse error - you can move it inline in the next line.

PR #166, opened 2020-05-12
172324177 sharing dialog uses the new share URLs

knowuh:
This PR should complete:

* [#172225454](https://www.pivotaltracker.com/story/show/172225454)Update CFM Shared Document Provider to use readWriteTokens from token service.
* [#172324177](https://www.pivotaltracker.com/story/show/172324177)Update the sharing dialog so it uses the new form of shared document URLs

These PR also begins start work on:
* [#172324185](https://www.pivotaltracker.com/story/show/172324185) Handle opening of legacy shared document URLs

Still waiting to resolve an issue with CORS on #172324185 above.


knowuh:
@dougmartin -- There is still some cleanup to do here.  I wanted to push a branch and get a PR up before our planning meeting.  If you would rather wait until its cleaner before reviewing it  — no problem.

src/code/client.js
```diff
@@ -408,11 +415,16 @@ class CloudFileManagerClient {
...
   openSharedContent(id) {
+    const { shareProvider } = this.state
     this._event('willOpenFile', {op: "openSharedContent"})
-    return (this.state.shareProvider != null ? this.state.shareProvider.loadSharedContent(id, (err, content, metadata) => {
-      if (err) { return this.alert(err, () => this.ready()) }
-      return this._fileOpened(content, metadata, {overwritable: false, openedContent: content.clone()})
-  }) : undefined)
+    if(shareProvider.loadSharedContent) {
+      shareProvider.loadSharedContent(id, (err, content, metadata) => {
+        if (err) {
```
dougmartin, line 422: Might be cleaner to use `if/else` and then remove the returns since this is inside a callback.

src/code/client.js
```diff
@@ -701,7 +713,16 @@ class CloudFileManagerClient {
   }
 
   isShared() {
-    return (this.state.currentContent != null ? this.state.currentContent.get("sharedDocumentId") : undefined) && !(this.state.currentContent != null ? this.state.currentContent.get("isUnshared") : undefined)
+    const currentContent = this.state?.currentContent
+    if(currentContent) {
+      const unshared = currentContent.get("usUnshared")
```
dougmartin, line 718: I think that is supposed to be `isUnshared` instead of `usUnshared`

src/code/providers/provider-interface.ts
```diff
@@ -369,6 +373,8 @@ class CloudContent {
     }
     if (this.content.sharedDocumentId != null) {
       shared.sharedDocumentId = this.content.sharedDocumentId }
+    if (this.content.sharedDocumentUrl != null) {
```
dougmartin, line 376: All of these night be cleaner as something like `shared.sharedDocumentUrl = this.content?.sharedDocumentUrl`.

src/code/providers/s3-provider.ts
```diff
@@ -90,31 +68,60 @@ class S3Provider extends ProviderInterface {
...
-          name: e.name,
-          type: e.type,
-          parent: metadata,
-          provider: this
+  load(metadata: CloudMetadata, callback: callbackSigLoad) {
+    let documentUrl = metadata.sharedDocumentUrl
+    // TODO: Why is there both a sharedDocumentId and sharedContentId?
+    const contentId = metadata.sharedContentId
+    const urlRegex = /^http/
+    if(contentId.match(urlRegex)) {
```
dougmartin, line 81: Can `contentId` be `undefined` here?

src/code/views/share-dialog-view.js
```diff
@@ -70,23 +70,27 @@ export default createReactClass({
...
+    if (shared) {
+      if(sharedDocumentUrl) {
+        // return sharedDocumentUrl
+        return `${this.props.client.getCurrentUrl()}#shared=${encodeURI(sharedDocumentUrl)}`
+      }
+      if(sharedDocumentId) {
+        if(useInternalLink) {
+          return `${this.props.client.getCurrentUrl()}#shared=${sharedDocumentId}`
+        }
+        const legacyUrl="https://token-service-files.concordqa.org/legacy-document-store"
```
dougmartin, line 89: Could this use the new `getLegacyUrl` method?

src/code/client.js
knowuh: Yikes -- thanks.

src/code/providers/s3-provider.ts
knowuh: Nice catch

knowuh:
Thanks for the review @dougmartin  

I think e4c57f6 should address most of your concerns.

PR #171, opened 2020-07-08
Copied over CODAP's CFM save retry logic [#172367547]

dougmartin:
This change allows the autosave to retry saving instead of disabling autosave after the first error.

It also adds a missing live-server dependency used in the package.json scripts and a missing dependency used to pull the PO editor strings.

The CODAP commit is 8c1843fb45643e641d3477ba48be18875e6cb62e.


src/code/client.js
```diff
@@ -561,24 +561,36 @@ class CloudFileManagerClient {
...
-        // disable autosave on save failure; clear "Saving..." message
-        metadata.autoSaveDisabled = true
-        this._setState({ metadata, saving: null })
+        this._setState({ metadata, saving: null });
         if (statusCode === 403) {
-          return this.confirmAuthorizeAndSave(stringContent, callback)
+          return this.confirmAuthorizeAndSave(stringContent, callback);
         } else {
-          return this.alert(err)
+          failures = this.state.failures
```
kswenson, line 570:
```suggestion
          failures = this.state.failures;
```
It appears that we're inconsistent on commas.

kswenson: Upon further review it appears that _not_ having semicolons is now the project default, so I'm going to go ahead and remove them here.

kswenson:
I'm going to go ahead and merge this because it will help with my `[CODAP-sync]` PRs.

PR #172, opened 2020-07-10
[CODAP-sync] Enable CSV import by URL parameter

kswenson:

Contains the contents of two commits which reference two PT stories.

---
[[#162736733]](https://www.pivotaltracker.com/story/show/162736733) Make it possible to launch CODAP with a CSV via URL parameter
* Remove JSON type in ajax request for URL Provider

[38129daf](https://github.com/concord-consortium/cloud-file-manager/commit/38129daf12e4ebfb8d6e603dce72a4b7d032099b) Cherry-picked from CODAP branch

---
[[#166523992]](https://www.pivotaltracker.com/story/show/166523992) Fix bug: Cannot open csv from URL param.

[efeb6a69](https://github.com/concord-consortium/cloud-file-manager/commit/efeb6a698a262c562d29914ddf26a10d29eac5c7) Cherry-picked from CODAP branch

### DIFF
--- a/src/code/client.js
+++ b/src/code/client.js
@@ -1076,7 +1076,7 @@ class CloudFileManagerClient {
     this._updateState(content, metadata, additionalState, hashParams)
     // add metadata contentType to event for CODAP to load via postmessage API (for SageModeler standalone)
     const contentType = metadata.mimeType || metadata.contentType
-    if (contentType) { eventData.metadata = {contentType} }
+    eventData.metadata = {contentType, url: metadata.url, filename: metadata.filename}
     return this._event('openedFile', eventData, (iError, iSharedMetadata) => {
       if (iError) { return this.alert(iError, () => this.ready()) }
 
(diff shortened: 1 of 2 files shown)


dougmartin:
@kswenson: This seems to be missing this change from the CODAP branch:

https://github.com/concord-consortium/cloud-file-manager/commit/38129daf12e4ebfb8d6e603dce72a4b7d032099b#diff-b6ee016e90c0aa59b58136df28150e8fR700

kswenson:
> Seems to be missing this change from the referenced CODAP branch: 38129da

That's actually intentional. Sorry if that wasn't clear. Reconstructing the sequence I think I inadvertently removed that line when merging `master` to `CODAP` at some point, which introduced the bug mentioned in the second PT story, which Jonathan then fixed with the second commit, which essentially does the same thing (sets the metadata) in a more sophisticated way, which I believe obviates the need for this particular change.

kswenson:
I'm going to merge this so I can test it alongside the `config-improvements` branch.

PR #173, opened 2020-07-10
[CODAP-sync] Readonly-provider supports relative urls

kswenson:

---
Modify readonly-provider to resolve relative urls against config URL.

[dbeba51a](https://github.com/concord-consortium/cloud-file-manager/commit/dbeba51a600871aee710e33070696daee40a2e4e) Cherry-picked from CODAP branch


ddamelin:
Looks good 👍

PR #174, opened 2020-07-10
[CODAP-sync] Check language of page before language of browser

kswenson:

---
[[#172561086]](https://www.pivotaltracker.com/story/show/172561086) Check language of page before language of browser

[b635a9fa](https://github.com/concord-consortium/cloud-file-manager/commit/b635a9faaf50c290e08434bac856efdb4b396dba) Cherry-picked from CODAP branch

### DIFF
--- a/src/code/utils/translate.js
+++ b/src/code/utils/translate.js
@@ -37,6 +37,14 @@ const getBaseLanguage = function(langKey) {
   return undefined
 }
 
+// use language of page, which is used by CODAP, with separate build for each language
+const getPageLanguage = function() {
+  const pageLang = document.documentElement.lang
+  return pageLang && (pageLang !== "unknown")
+          ? pageLang
+          : undefined
+}
+
 const getFirstBrowserLanguage = function() {
   const nav = window.navigator
   const languages = nav ? (nav.languages || []).concat([nav.language, nav.browserLanguage, nav.systemLanguage, nav.userLanguage]) : []
@@ -54,7 +62,7 @@ languageFiles.forEach(function(lang) {
   if (baseLang) { return translations[baseLang] = lang.contents }
 })
 
-const lang = urlParams.lang || getFirstBrowserLanguage()
+const lang = urlParams.lang || getPageLanguage() || getFirstBrowserLanguage()
 const baseLang = getBaseLanguage(lang || '')
 const defaultLang = lang && translations[lang] ? lang : baseLang && translations[baseLang] ? baseLang : "en"
 


kswenson:
@dougmartin Were there particular changes you're requesting for this PR?

dougmartin:
@kswenson nope - I just hit the wrong radio button approving the changes :)

ddamelin:
I'm curious about why we want to try to get the page language.

In SageModeler the share link will not specify the lang param and so you would normally get a link that would then display the document in the default browser language. However, in CODAP the URL path has a language component (i.e. https://codap.concord.org/releases/latest/static/dg/**es**/cert/#shared=159470 for Spanish) does include reference to the Spanish build of CODAP. If that causes the index.html that is loaded to have a page with the document.documentElement.lang attribute set to "es" then share links will show up in Spanish everywhere using this code regardless of browser default language. (I just checked production CODAP and this seems to be the case.) However, in past discussions I thought we decided that it would be better that if you were sharing a CODAP document that the CODAP UI for that doc show up in the browser's native language, rather than using the page language. Can someone describe why we diverted from that original guidance? @jsandoe @bfinzer

kswenson:
What if share links stripped the language from the URL (e.g. https://codap.concord.org/releases/latest#shared=159470) or perhaps (with more work) gave authors the option to share the URL with or without the language code?

jsandoe:
@ddamelin, perhaps there are two separate issues. I think the fix represented here is to make it so the CFM will respect the page's language, even when the browser's language is different. If I, using my English language browser, select Spanish from the dropdown on the right of the CFM, I am changing the page's language, but not the browser's language. It makes sense to have that setting to have primacy. The CFM could get the language from the document or from a CODAP configuration parameter, but the result would be the same. I think in the sage-modeler-site instance, the page language would, absent other factors, correspond to the browser language. 

The second issue is, I think, should CODAP generate shared links that dictate a language for the page to open in, ignoring the browser language? I agree that it shouldn't. Perhaps CODAP could configure CFM with a base URL for sharing. In other words, CODAP could specify that the sharing base url is "https://codap.concord.org/app", and the CFM would generate shared links like "https://codap.concord.org/app?shared=12834197".  I think this could be handled as a separate story, though.

ddamelin:
@jsandoe I see what you mean. Looks like the behavior here will be as expected for SageModeler (with the app UI loading in the browser default language), so this should be safe to merge. I still don't understand why it is needed. In your scenario, where the user has English as default and selects Spanish from the language menu, we show the UI in Spanish already, even without this change, so I'm not sure what this change accomplishes. If you were to save the doc to google drive and then re-open it at a later time, CODAP, upon launching, will be in the browser default language, and opening the file will then open it in the CODAP current language setting. So, unless we are saving the language with the file somehow, I don't understand how this code will ever be executed. (Except in the case I described above via the CODAP share link, but it seems  you agree that should change, so again not sure why this code is necessary.) Anyway, since I don't see it causing issues I'm fine approving this change.

kswenson:
>I still don't understand why it is needed. In your scenario, where the user has English as default and selects Spanish from the language menu, we show the UI in Spanish already, even without this change

As you say, this is the desired behavior but it's only the current behavior in CODAP because there is code on the `codap` branch (the change in this PR) to prefer the page language to the browser language. Without that change, CODAP strings are displayed in the page language but CFM strings are displayed in the browser default language. The point of this PR is to unify the code on the `codap`/`master` branches.

ddamelin:
> As you say, this is the desired behavior but it's only the current behavior in CODAP because there is code on the `codap` branch (the change in this PR) to prefer the page language to the browser language. Without that change, CODAP strings are displayed in the page language but CFM strings are displayed in the browser default language. The point of this PR is to unify the code on the `codap`/`master` branches.

Now I got it. Thanks.

PR #175, opened 2020-07-10
[CODAP-sync] Synchronize languages

kswenson:

---
This PR adds Japanese and Simplified Chinese which have recently been added to the CODAP branch. The last step is to pull the most recent strings from POEditor once I have appropriate credentials to do so.


src/code/utils/translate.js
```diff
@@ -10,24 +10,28 @@
...
+  {key: 'de',    contents: de},     // German
+  {key: 'el',    contents: el},     // Greek
+  {key: 'en-US', contents: enUS},   // US English
+  {key: 'es',    contents: es},     // Spanish
+  {key: 'he',    contents: he},     // Hebrew
+  {key: 'ja' ,   contents: ja},     // Japanese
+  {key: 'nb',    contents: nb},     // Norwegian Bokmål
+  {key: 'nn',    contents: nn},     // Norwegian Nynorsk
+  {key: 'tr',    contents: tr},     // Turkish
+  {key: 'zh-tw', contents: zhHans}, // Traditional Chinese (Taiwan)
```
dougmartin, line 33: Due to brower's not returning zh-hans for simplified Chinese but returning zh-tw for traditional we defaulted to the following in sagemodeler (and I thought we had done the same in CFM):

```
  "zh":      require("./lang/zh-HANS.json"), // Chinese (Simplified)
  "zh-TW":   require("./lang/zh-TW.json"),   // Chinese (Traditional)
```

(https://github.com/concord-consortium/sage-modeler-site/blob/master/src/code/utils/translate.ts#L12-L13)

this lets us default to simplfied Chinese for all zh readers except for those that exactly specify zh-tw.

More info in these two threads on Slack:

- https://concord-consortium.slack.com/archives/C09JNV371/p1591364967116300
- https://concord-consortium.slack.com/archives/C09JNV371/p1591372571128300

kswenson: Thx for the pointers -- I'll incorporate what I can from those links and let you know when it's ready for re-review.

ddamelin:
Looks good 👍

PR #178, opened 2020-07-10
Add information about caching to Shared File Update dialog

pjanik:
https://www.pivotaltracker.com/story/show/173734184

I've modified interfaces/options, so it's easier to ensure that message in the dialog matches reality.


pjanik:
Thanks, I've added changes to en-US-master too.

pjanik:
My last commit addresses the only review comment, so I'll go ahead and merge, as my current task is touching the same files.

PR #181, opened 2020-07-13
Remove CFMOrigin from LARA share dialog

pjanik:
PT story provides more detail why it's removed:
https://www.pivotaltracker.com/story/show/173755317
If necessary, we can edit the share URL manually.

I've also removed UI related to launch page that isn't supported anymore.


dougmartin:
Code looks good to me too however I'm also not very familiar with the new S3 code that Noah added to the CFM so I'm not 100% confident why it was added to the dialog in the first place.

pjanik:
I'm in the same boat. But I think it was useful for testing as PT story describes. Although this doesn't work correctly for SageModeler, and we can still add these params manually if we need to. Long term, I think this could be re-added, maybe just hidden/shown based on some URL param. But we'd need to think how to handle CODAP vs SageModeler too.

PR #183, opened 2020-07-14
Handle update of legacy shared docs

pjanik:
[#173798525]

Legacy documents don't have the correct document ID and readWriteToken. These things have to be generated using the data we have available. document-store migration script does the same thing, and code in both places have to match:
https://github.com/concord-consortium/document-store/blob/master/token-service-migration/index.js#L115-L126

It doesn't fully fix update of shared docs, as now we have this issue:
Screenshot 2020-07-14 at 20 15 53
but it seems to be token-service related.


src/code/providers/s3-share-provider.ts
```diff
@@ -101,19 +102,39 @@ class S3ShareProvider implements IShareProvider  {
...
+      if (isLegacyDocument) {
+        // This logic is based on logic in the document store migration script:
+        // https://github.com/concord-consortium/document-store/blob/master/token-service-migration/index.js#L115-L126
+        documentID = accessKeys.readOnly
+        if (!documentID) {
+          // document-store migration does the same thing if readOnly key is missing.
+          documentID = sha256(readWriteToken)
+        }
+        // Again, follow document-store migration code.
+        readWriteToken = "read-write-token:doc-store-imported:" + readWriteToken
```
dougmartin, line 135: Based on your Slack comment this should be something like:

`("read-write-token:doc-store-imported:" + readWriteToken).substr(0, 64)`

or

`("read-write-token:imported:" + readWriteToken).substr(0, 64)`

pjanik: I don't think this error is about that. readWriteToken can't be truncated, as that's the key that token-service will compare with one saved in Firestore.

I believe the token-service error is coming from:
https://github.com/concord-consortium/token-service/blob/master/functions/src/base-resource-object.ts#L432
So I think that truncation needs to happen there.

PR #187, opened 2020-07-20
Fix linked interactives, LARA Embed URL construction

knowuh:
This fixes the bug with linked interactive URLs in the LARA provider using document-server (numeric) ids.

It also fixes a bug with overly complex links being generated by `getShareLink` in `share-dialog-view.js`

For the migration task:

The `new_url` format for the migration of authored interactives is this:

https://<CFM_URL>/autolaunch/autolaunch.html
 ?documentId=<url-encoded https://models-resources.concord.org/Flegacy-document-store/<LEGACY_ID>>
 &server=<CODAP_URL> (we had been including cfmBaseUrl, but I don't think we need to)
 &scaling

eg:

```JSON
{
    "id": 207249,
    "old_url": "https://document-store.concord.org/v2/documents/16046/autolaunch?server=https%3A%2F%2Fcodap.concord.org%2Freleases%2Flatest%2F&buttonText=Launch",
    "new_url": "https://cloud-file-manager.concord.org/branch/master/autolaunch/autolaunch.html?documentId=https%3A%2F%2Fmodels-resources.concord.org%2Flegacy-document-store%2F16046&scaling=true&server=https%3A%2F%2Fcodap.concord.org%2Freleases%2Fstaging%2F"
}
```

This commit also allows live-server to bind to 'localhost' for an address.

### DIFF
--- a/src/code/views/share-dialog-view.js
+++ b/src/code/views/share-dialog-view.js
@@ -77,8 +77,8 @@ export default createReactClass({
     return client.state?.currentContent?.get("sharedDocumentId")
   },
 
-  // New S3ShareProvider saves shareDocumentUrl
-  getShareLink() {
+  // Get a documentID, or going forward, always return a URL to a document
+  getShareUrl() {
     const { client } = this.props
     const shared = client.isShared()
     const sharedDocumentUrl = client.state?.currentContent?.get("sharedDocumentUrl")
@@ -87,9 +87,20 @@ export default createReactClass({
       if(sharedDocumentUrl) {
         return sharedDocumentUrl
       }
+      // 2020-07-20 NP: We still need to support simple numeric IDs for now
+      // in the future we can convert w getLegacyUrl(sharedDocumentId)
       if(sharedDocumentId) {
-        return getLegacyUrl(sharedDocumentId)
+        return sharedDocumentId
       }
+    }
+    return null
+  },
+
+  // Return a sharable link to this document (includes the app CODAP)
+  getShareLink() {
+    const shareRef = this.getShareUrl()
+    if (shareRef) {
+      return `${this.props.client.getCurrentUrl()}#shared=${encodeURI(shareRef)}`
     }
     return null
   },
@@ -110,7 +121,7 @@ export default createReactClass({
     //  `<AutoLaunchPageUrl>?documentId=<XX>&server=<CODAP SERVER>&scaling`
     // <AutoLaunchPageUrl> expects a `documentId` param (was a path in DocStore)
     //    and a `server` param, that points usually to some CODAP release.
-    // <CODAP SERVER> can have url encoded params too such as `cfmBaseUrl=`
+    // <CODAP SERVER> can have url encoded params too such as `FcfmBaseUrl=`
     // where URL is the url to the /js folder for CFM
     //
     // It will point to SPA hosted in this repo -- NP 2020-06-29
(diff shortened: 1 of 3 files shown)


knowuh:
@pjanik @scytacki 

Not much code changed. 😀

I feel pretty good about the changes, I tested several linked interactive scenarios, and everything seemed good.

I appreciate your help and patience on this.

src/code/views/share-dialog-view.js
```diff
@@ -83,15 +83,11 @@ export default createReactClass({
     const shared = client.isShared()
     const sharedDocumentUrl = client.state?.currentContent?.get("sharedDocumentUrl")
     const sharedDocumentId = this.getSharedDocumentId()
-    const useInternalLink = true
     if (shared) {
       if(sharedDocumentUrl) {
-        return `${this.props.client.getCurrentUrl()}#shared=${encodeURI(sharedDocumentUrl)}`
+        return sharedDocumentUrl
```
scytacki, line 88: I think this approach breaks the normal share links. I tried it here: https://codap.concord.org/releases/latest/static/dg/en/cert/index.html?cfmBaseUrl=https://cloud-file-manager.concord.org/branch/simplify-share-urls/js
When I share a document and I look at the 'Link' tab of the sharing dialog I see a URL of:
https://cfm-shared.concord.org/NcSu07SspsIyKNpZE5Ge/file.json
That link won't work if someone sends it to someone else.

So I think what is needed is that the getShareLink needs to stay mostly the same. And the code in `getLARA()` needs to change so it uses just the sharedDocumentUrl

pjanik:
>  (we had been including cfmBaseUrl, but I don't think we need to)

I guess we need to do that before we update CFM production and CODAP production. So, cfmBaseUrl will be included only on LARA staging. When we do final migration on production, it won't be included.

knowuh:
@scytacki  thanks for pointing out stand-alone-sharing.

I believe it's covered now -- standalone share links and LARA links seem to both work after  136d722

The comments in 136d722 discuss that while we are in the process of changing things, we need to manually modify the share URLS so everything is using the correct version of the CFM.  This is why I had originally included a CFMBaseUrl field in the share-dialog-view.  But this should work.

pjanik:
> The comments in 136d722 discuss that while we are in the process of changing things, we need to manually modify the share URLS so everything is using the correct version of the CFM. This is why I had originally included a CFMBaseUrl field in the share-dialog-view. But this should work.

To clarify this comment even more:

> So once you get a sharable url, you will need to add a query param to the CODAP host.  eg, for localhost:
> `https://codap.concord.org/releases/latest/static/dg/en/cert/index.html?cfmBaseUrl=https://localhost:8888/js#shared=https://cfm-shared.concord.org/mXlxnxoQsSZNEsrzHbaL/file.json`

This is correct for CODAP. Although for SageModeler, we would need cfmBaseUrl, sage:cfmBaseUrl, and codap:cfmBaseUrl instead with a bit different values. I guess that's the main reason why @scytacki thought it CFMOrgin should be removed from Share Dialog - it wasn't handling this case and it was always overwriting the cfmBaseUrl.

> Similarly the LARA links need to be manually updated to use the correct base URL for the CFM.  By default they look like:
> `https://cloud-file-manager.concord.org/autolaunch/autolaunch.html?documentId=https%3A%2F%2Fcfm-shared.concord.org%2FmXlxnxoQsSZNEsrzHbaL%2Ffile.json&server=https%3A%2F%2Fcodap.concord.org%2Freleases%2Flatest%2F&scaling`
> We would need to replace `cloud-file-manager.concord.org`  with whatever release we are using.

Yes, but we also need to add cfmBaseUrl (or cfmBaseUrl, sage:cfmBaseUrl, and codap:cfmBaseUrl) in `server` URL param.

The migration script should take care of these things, but just pointing this out in case you're manually updating URLs and testing.

scytacki:
Looks great to me.

Piotr is right about the cfmBaseUrl and issues with Sage. 
I'm going to merge this in to keep things moving along.

PR #188, opened 2020-07-22
Fix JQuery JSON PUT request to Docstore on S3 Copy

knowuh:
JQuery was converting JSON documents into multipart form documents, and breaking Rack's upper limit on number of parameter keys.

[#173950259]

### DIFF
--- a/src/code/providers/lara-provider.js
+++ b/src/code/providers/lara-provider.js
@@ -497,7 +497,10 @@ class LaraProvider extends ProviderInterface {
             dataType: 'json',
             contentType: 'application/json', // Document Store requires JSON currently
             processData: false, // https://api.jquery.com/jquery.ajax/
-            data: dataJson,
+            data: pako.deflate(dataJson),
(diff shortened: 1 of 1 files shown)


src/code/providers/lara-provider.js
```diff
@@ -495,6 +495,8 @@ class LaraProvider extends ProviderInterface {
           return $.ajax({
             type: method,
             dataType: 'json',
+            contentType: 'application/json', // Document Store requires JSON currently
+            processData: false, // https://api.jquery.com/jquery.ajax/
             data: dataJson,
```
scytacki, line 500: Wouldn't it be worthwhile to compress this data just like the other upload code is doing?

knowuh: Sorry, I forgot about that.

b050dec re-adds the compression.  


Also I had to read the pako documentation, and then re-test everything, because I thought for sure `pako.deflate()` would decompress content.

PR #196, opened 2020-10-16
Changed to always change the Google Drive v2 downfile file location

dougmartin:
The v2 api is not returning the correct domain for the download file url making opens fail.  Previously we fell back to changing the domain when the initial domain load failed but a recent change in the behavior of the v2 api causes the error handler not to be called.

### DIFF
--- a/src/code/providers/google-drive-provider.js
+++ b/src/code/providers/google-drive-provider.js
@@ -71,9 +71,9 @@ const GoogleDriveAuthorizationDialog = createReactClassFactory({
 
 class GoogleDriveProvider extends ProviderInterface {
   static initClass() {
-  
+
     this.Name = 'googleDrive'
-  
+
     // aliases for boolean parameter to authorize
     this.IMMEDIATE = true
     this.SHOW_POPUP = false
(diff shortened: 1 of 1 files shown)


pjanik (approved):
Changed lines make sense to me. I might not be the best reviewer though, as I'm missing context - I'm not familiar with google drive API and all the issues.

src/code/providers/google-drive-provider.js
```diff
@@ -330,22 +330,22 @@ class GoogleDriveProvider extends ProviderInterface {
             }
           })
         }
-        var download = (url, fallback) => {
+        var download = (url) => {
```
knowuh, line 333: On line 319 we ask for the `file.downloadUrl`  

A response in the bug report ( https://issuetracker.google.com/issues/149891169#comment7 ), suggests we should be asking for webContentLink instead. I am not sure if we tried that, or if it would work in our case, but it would be great if we could use the API cleanly.

knowuh (approved):
This looks good to me. 

I made one inline comment about an alternative we could try if we had time / energy.

src/code/providers/google-drive-provider.js
dougmartin: The webContentLink doesn't have the right mimeType or CORs headers to load the file into the app.  Its really a download link to save the file from your browser to your local machine.

dougmartin:
@knowuh I made a couple of changes after your review to clean remove the unneeded boolean parameter on the call to download and to add back the check for the protocol + domain instead of just the domain for extra precaution in case the path also had the domain for some reason.  Could you review those changes?

knowuh (approved):
Looks good to me. Thanks

PR #199, opened 2020-11-03
[#175445858] CFM sets wrong Chinese variant...

jsandoe:
...when Traditional writing system selected.

Sproutcore sets the language in the DocumentElement. It lower cases the name (it expects it to be language only). The CFM determines locale by checking, among other things, the DocumentElement setting. It was failing to match with 'zh-TW'. We adjust the CFM to do a case insensitive mapping.


kswenson (approved):
👍 LGTM

PR #201, opened 2021-02-08
[#176443113] Fix bug: build corrupts binaries...

jsandoe:
... for developers with 'js' in their user names.

Also increment version number.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "cloud-file-manager",
-  "version": "1.1.1",
+  "version": "1.1.3",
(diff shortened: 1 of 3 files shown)


kswenson (approved):
👍 Looks good, although I would be remiss in not pointing out that simply changing your name would have required fewer code changes. 😉

jsandoe:
Maybe true that changing my name would have helped, but Charles S Smith would have been in trouble, if he became a Concord Developer.

kswenson:
Only if he used his middle initial in his username. 😉

PR #207, opened 2021-04-30
Fixed broken language translation

dougmartin:
An earlier merged commit lowercased the language key before filling the translation tables but the language key was not then also lowercased when a translation was looked up causing en-US and zh-TW to fail.


dougmartin:
I'm going to merge this without review as production currently shows the translation markers instead of the translated text for English and Chinese after the production update a few minutes ago.  I've tested it locally.

PR #213, opened 2021-06-15
Update dependencies & lint configuration; improve some types

kswenson:
- update dependencies
- update eslint configuration
- fix lint warnings/errors
- improve some provider typings

There's more to do, but this seemed like a reasonable place to break for the first PR.


src/code/views/share-dialog-view.js
```diff
@@ -20,7 +20,7 @@ import {ShareLoadingView} from './share-loading-view'
 // of the react function, "tr".
 import translate  from '../utils/translate'
 import socialIcons  from 'svg-social-icons/lib/icons.json'
-import { getLegacyUrl } from '../utils/s3-share-provider-token-service-helper'
+// import { getLegacyUrl } from '../utils/s3-share-provider-token-service-helper'
```
dougmartin, line 23: Shouldn't you just delete this import line?

dougmartin (approved):
Looks good :+1:

src/code/providers/provider-interface.ts
```diff
@@ -468,8 +457,7 @@ class ProviderInterface implements IProviderInterfaceOpts {
       for (let extension of CloudMetadata.ReadableExtensions) {
         if (name.substr(-extension.length) === extension) { return true }
         if (extension === "") {
-          // TODO:  Eslint doesn't like this bitwise not operator.
-          if (!~name.indexOf(".")) { return true }
+          if (name.indexOf(".") === -1) { return true }
```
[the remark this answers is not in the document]
kswenson: I prefer `>= 0` for the positive test ("is a valid index"). The `!== -1` variant ("is not the magic sentinel value that indicates failure") always seems like an overly complicated way of testing for success by double-negative. The case for the negative test (`=== -1` vs. `< 0` vs. `!(_ >= 0)`) seems less clear to me as there isn't as obvious an advantage in simplicity, but the former seems somewhat more conventional than the others.

src/code/views/share-dialog-view.js
kswenson: I thought perhaps its presence might be meaningful to someone ("Hmmm, I wonder why that's not used?") but that's probably not a good reason to keep it around. I'll remove it.

PR #214, opened 2021-06-22
TypeScript migration [#175170529]

kswenson:
@dougmartin As mentioned in standup you may need to pack a lunch for this one. I tried to keep the changes to logical commits, but things started to bleed together at some point and there are more commits than I originally anticipated, so you may just want to ignore the commits and look at the diffs all at once. I added comments to indicate some of the places where behavior was changed, usually because the previous behavior was somewhere on the spectrum from questionable to obviously wrong.

The TypeScript migration generated a bunch of issue comments that have since been removed/addressed. Here's a table that I created to track progress along the way.

# TypeScript Conversion (June 2021)

## Issues flagged by ts-migrate

|Error|Fixed|Description|Original Count|Current Count|
|-----|-----|-----------|-------------:|------------:|
|2304|:white_check_mark:|Cannot find name __.|78|0|
|2307|:white_check_mark:|Cannot find module __ or its corresponding type declarations.|12|0|
|2322|:white_check_mark:|Type __ is not assignable to type __.|4|0|
|2339|:white_check_mark:|Property __ does not exist on type 'Readonly<{}>'|2|0|
|2345|:white_check_mark:|Argument of type __ is not assignable to parameter of type __.|3|0|
|2349|:white_check_mark:|This expression is not callable.|1|0|
|2362|:white_check_mark:|Type __ is not assignable to type __.|1|0|
|2376|:white_check_mark:|A 'super' call must be the first statement in the constructor when a class contains initialized properties, parameter properties, or private identifiers.|1|0|
|2416|:white_check_mark:|Property __ in type __ is not assignable to the same property in base type __.|2|0|
|2551|:white_check_mark:|Property __ does not exist on type __.|1|0|
|2554|:white_check_mark:|Expected __ arguments, but got __.|168|0|
|2686|:white_check_mark:|__ refers to a UMD global, but the current file is a module.|18|0|
|2732|:white_check_mark:|Cannot find module __. Consider using '--resolveJsonModule' to import module with '.json' extension.|1|0|
|7006|:white_check_mark:|Parameter __ implicitly has an 'any' type.|34|0|
|7010|:white_check_mark:|__, which lacks return-type annotation, implicitly has an __ return type.|2|0|
|7011|:white_check_mark:|Function expression, which lacks return-type annotation, implicitly has an 'any' return type.|2|0|
|7016|:white_check_mark:|Could not find a declaration file for module __|16|0|
|7018|:white_check_mark:|Object literal's property __ implicitly has an 'any' type.|3|0|
|7022|:white_check_mark:|__ implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.|1|0|
|7023|:white_check_mark:|__ implicitly has return type 'any' because it does not have a return type annotation and is referenced directly or indirectly in one of its return expressions.|2|0|
|7024|:white_check_mark:|__ implicitly has return type 'any' because it does not have a return type annotation and is referenced directly or indirectly in one of its return expressions.|1|0|
|7053|:white_check_mark:|Element implicitly has an 'any' type because expression of type 'any' can't be used to index type '{}'|15|0|
|||**Grand Total**|359|0|


build-support/build-opts.js
```diff
@@ -23,7 +23,12 @@ const replacementStrings = {
     }
   ],
   css: [],
-  js: []
+  js: [
+    {
+      search: /__PACKAGE_VERSION__/g,
```
kswenson, line 28: The version replacement in the CFM metadata appears to have been broken since the gulp => webpack transition.

live-server-ssl-config.js
```diff
@@ -1,16 +1,22 @@
...
+ * 4. Make the cert files: `mkcert -cert-file localhost.pem -key-file localhost.key localhost 127.0.0.1 ::1`
+ * 5. Run `npm run start:secure` to run `webpack-dev-server` in development mode with hot module replacement
+ *
+ **/
 
 module.exports = {
-    cert: fs.readFileSync(__dirname + "/LocalhostCertificates/localhost.crt"),
-    key: fs.readFileSync(__dirname + "/LocalhostCertificates/localhost.key")
+    cert: fs.readFileSync(path.resolve(os.homedir(), ".localhost-ssl/localhost.pem")),
+    key: fs.readFileSync(path.resolve(os.homedir(), ".localhost-ssl/localhost.key"))
```
kswenson, line 21: I was originally going to add this as a separate option so that both the local directory and home directory options were supported, but in conversation with Scott he suggested that we should just move projects to the new convention.

src/assets/examples/enable-demo-extension.html
```diff
@@ -12,7 +12,7 @@
       var options = {
         app: "example-app/index.html",
         mimeType: "text/plain",
-        extension: ".demo",
+        extension: "demo",
```
kswenson, line 15: With the "." here, you get `foo..demo`

src/code/autolaunch/autolaunch.ts
```diff
@@ -24,14 +24,14 @@ function getURLParam (name) {
...
-function codap_v2_link(codapServer) {
+function codap_v2_link(codapServer: any) {
   const defaultCodapUrl = "https://codap.concord.org/releases/latest/"
   const documentServer = "https://document-store.concord.org/"
   const extraData = {
     documentServer
   }
   const baseUrl = codapServer || defaultCodapUrl
-  const {url, query, fragmentIdentifier} = queryString.parseUrl(baseUrl, {fragmentIdentifier: true})
+  const {url, query, fragmentIdentifier} = queryString.parseUrl(baseUrl, {parseFragmentIdentifier: true})
```
kswenson, line 34: This was a bug identified by the TypeScript conversion: `fragmentIdentifier` is not a valid option to `parseUrl()`.

src/code/autolaunch/autolaunch.ts
```diff
@@ -171,8 +171,8 @@ export default function autolaunchInteractive() {
     var launchParams = {url: interactiveData.interactiveStateUrl, source: documentId, collaboratorUrls: interactiveData.collaboratorUrls}
     // If there is a linked state and no interactive state then change the source document to point to the linked recordid and add the access key.
     if (stateValid(linkedState) && !interactiveStateAvailable) {
-      launchParams.source = linkedState.docStore.recordid
-      launchParams.readOnlyKey = linkedState.docStore.accessKeys.readOnly
+      launchParams.source = linkedState.docStore.recordid;
```
kswenson, line 174: The inconsistency of the no-semicolons convention is one of the things that annoys me about it.

src/code/client.ts
```diff
@@ -430,40 +465,44 @@ class CloudFileManagerClient {
...
   // must be called as a result of user action (e.g. click) to avoid popup blockers
   parseUrlAuthorizeAndOpen() {
-    if ((this.appOptions.hashParams != null ? this.appOptions.hashParams.fileParams : undefined) != null) {
+    if (this.appOptions.hashParams?.fileParams != null) {
       const [providerName, providerParams] = this.appOptions.hashParams.fileParams.split(':')
       const provider = this.providers[providerName]
       if (provider) {
         return provider.authorize(() => {
-          return this.openProviderFile(providerName(providerParams))
+          return this.openProviderFile(providerName, providerParams)
```
kswenson, line 473: Another bug made apparent by the TypeScript conversion: `providerName` is a string and so can't be called like a function. This presumably dates back to the CoffeeScript era where the code probably looked something like: `@openProviderFile providerName providerParams`.

src/code/client.ts
```diff
@@ -1001,23 +1044,23 @@ class CloudFileManagerClient {
           }
         }
         if ((this.state.metadata != null ? this.state.metadata.provider : undefined) || this.autoProvider('save')) {
-          return this.save(err => postSave())
+          return this.save((err: string | null) => postSave(err))
```
kswenson, line 1047: Another apparent bug -- not passing `err` to the postSave callback.

src/code/utils/is-array.ts
```diff
@@ -0,0 +1 @@
+export default (value: any) => Array.isArray(value)
```
kswenson, line 1: The previous implementation of this function appeared to be buggy as well:
```javascript
export default value => Array.isArray(value || {}.toString.call(value === '[object Array]'))
```

src/code/views/file-dialog-tab-view.ts
```diff
@@ -93,16 +96,15 @@ const FileList = createReactClassFactory({
     if (isSubFolder) {
       list.push((div({key: 'parent', className: 'selectable', onClick: this.parentSelected}, (italic({className: 'icon-paletteArrow-collapse'})), this.props.folder.name)))
     }
-    for (i = 0; i < this.props.list.length; i++) {
+    for (let i = 0; i < this.props.list.length; i++) {
```
kswenson, line 99: As far as I can tell, this was previously modifying the `i` imported from `ReactDOMFactories` above. :scream:

src/code/providers/google-drive-provider.ts
```diff
@@ -226,50 +248,50 @@ class GoogleDriveProvider extends ProviderInterface {
...
       const request = gapi.client.drive.files["delete"]({
         fileId: metadata.providerData.id})
-      return request.execute(result => typeof callback === 'function' ? callback((result != null ? result.error : undefined) || null) : undefined)
+      return request.execute((result: any) => callback?.(result?.error))
     })
   }
 
-  rename(metadata, newName, callback) {
+  rename(metadata: CloudMetadata, newName: string, callback?: (err: string | null, metadata?: CloudMetadata) => void) {
     return this._waitForGAPILoad().then(() => {
```
kswenson, line 260: Not directly related to this change, but the _waitForGAPILoad() function looks like this:
```typescript
  _waitForGAPILoad() {
    return new Promise((resolve, reject) => {
      const script = document.createElement('script')
      script.src = "https://apis.google.com/js/api.js"
      script.onload = () => {
        gapi.load("client:auth2", () => {
          gapi.client.init({
            apiKey: this.apiKey,
            clientId: this.clientId,
            discoveryDocs: ["https://www.googleapis.com/discovery/v1/apis/drive/v3/rest"],
            scope: this.scopes
          })
          .then(resolve)
          .catch(reject)  // eslint-disable-line @typescript-eslint/dot-notation
        })
      }
      document.head.appendChild(script)
    })
  }
```
Unless I'm missing something, that appears to be loading the API every time it's called. Given that this function is called by nearly every other function in this module, that seems to suggest that we're loading the API more than necessary. I would have expected to see a check for whether the API is already loaded and an immediate return if detected.

src/code/providers/document-store-provider.ts
```diff
@@ -136,19 +147,19 @@ class DocumentStoreProvider extends ProviderInterface {
     }
   }
 
-  authorize(completionCallback) {
+  authorize(completionCallback: any) {
```
dougmartin, line 150: `completionCallback: () => void`

src/code/providers/document-store-provider.ts
```diff
@@ -173,14 +184,14 @@ class DocumentStoreProvider extends ProviderInterface {
...
-  _showLoginWindow(completionCallback) {
+  _showLoginWindow(completionCallback: () => void) {
     if (this._loginWindow && !this._loginWindow.closed) {
       this._loginWindow.focus()
     } else {
 
-      const computeScreenLocation = function(w, h) {
-        const screenLeft = window.screenLeft || screen.left
-        const screenTop  = window.screenTop  || screen.top
+      const computeScreenLocation = function(w: any, h: any) {
```
dougmartin, line 192: `w: number, h: number`

src/code/providers/google-drive-provider.ts
dougmartin: I think you are correct.  We should have a class member to act as a singleton holding the Promise and do something like:

```
return this.loadPromise || (this.loadPromise = return new Promise(...
```

src/code/providers/lara-provider.ts
```diff
@@ -87,18 +124,18 @@ class LaraProvider extends ProviderInterface {
...
   }
 
   // don't show in provider open/save dialogs
-  filterTabComponent(capability, defaultComponent) {
+  filterTabComponent(capability: ECapabilities, defaultComponent: React.Component): React.Component | null {
     return null
   }
 
-  extractRawDataFromRunState(runState) {
+  extractRawDataFromRunState(runState: any) {
```
dougmartin, line 138: I think this would be better typed as `runState: string|object`

src/code/views/file-dialog-tab-view.ts
dougmartin: You have heard of iterative italics right? :)

dougmartin (approved):
Looks good :+1: 

I had some minor typing comments you can feel free to ignore.  I kind of ran out of stream adding them towards the end.  I should have packed a bigger sandwich.

kswenson:
>I had some minor typing comments you can feel free to ignore. I kind of ran out of stream adding them towards the end. I should have packed a bigger sandwich.

Thx for the suggestions. I should have mentioned at the outset that I pretty much ran out of steam myself after spending over a week on it and figured that it made more sense to find a coherent stopping place than to keep chasing down every last `any`. That said, some of the areas you pointed out are ones that I consciously decided to put off for another day and others I never really got to, so I'll see which of your suggestions can be implemented without too much difficulty and put in one more commit before merging.

src/code/app-options.ts
```diff
@@ -0,0 +1,124 @@
...
+  mimeType?: string;
+  // note different capitalization from CFMAppOptions
+  readableMimetypes?: string[];
+}
+
+export interface CFMReadOnlyProviderOptions extends CFMBaseProviderOptions {
+  src: string;
+  alphabetize: boolean;
+  json: any;
+  jsonCallback: (callback: (err: string | null, json: any) => void) => void;
```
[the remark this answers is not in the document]
kswenson: This one I think would require further investigation into the nested JSON structure used to represent a folder hierarchy so I'm inclined to leave this one for another day.

src/code/app-options.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+export interface CFMPatchProviderOptions extends CFMBaseProviderOptions {
+  patch: boolean;
+  patchObjectHash: (obj: any) => string;
```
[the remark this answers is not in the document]
kswenson: I thought about it but I'm not sure that it can't be a simple string so I decided to leave it for later.

src/code/autolaunch/autolaunch.ts
```diff
@@ -188,7 +188,7 @@ export default function autolaunchInteractive() {
     var iframeCanAutosave = false
     var iframeLoaded = function () {
       $(window).on('message', function (e) {
```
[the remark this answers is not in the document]
kswenson: It's more complicated because it's jQuery's wrapper for the MessageEvent, which is part of a deprecated API....

src/code/client.ts
```diff
@@ -24,30 +26,48 @@ import LocalFileProvider  from './providers/local-file-provider'
...
     this.callback = callback
     if (state == null) { state = {} }
     this.state = state
     CLOUDFILEMANAGER_EVENT_ID++
-    /** @type {number} */
     this.id = CLOUDFILEMANAGER_EVENT_ID
   }
 
-  postMessage(iframe) {
+  postMessage(iframe: any) {
```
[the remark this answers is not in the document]
kswenson: You would think so, but `Property 'postMessage' does not exist on type 'HTMLIFrameElement'.ts(2339)`.

src/code/client.ts
```diff
@@ -80,9 +113,11 @@ class CloudFileManagerClient {
     if (this.appOptions.wrapFileContent == null) { this.appOptions.wrapFileContent = true }
     CloudContent.wrapFileContent = this.appOptions.wrapFileContent
 
+    type ProviderClass = any;
```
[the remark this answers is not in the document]
kswenson: I thought the same thing at one point, but at runtime `ProviderClass` is actually the class of the provider to be created, not an instance of `ProviderInstance`, i.e. when we call `new Provider` in line 175 the `Provider` must be the actual class we're trying to instantiate.

src/code/client.ts
```diff
@@ -80,9 +113,11 @@ class CloudFileManagerClient {
...
+    const allProviders: Record<string, ProviderClass> = {}
+
     // Determine the available providers. Note that order in the list can
     // be significant in provider searches (e.g. @autoProvider).
-    const allProviders = {}
     const providerList = [
```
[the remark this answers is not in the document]
kswenson: Same reason as the previous comment -- these are provider classes not provider instances.

src/code/providers/lara-provider.ts
kswenson: The runState was one of those things that seemed to warrant a more thorough investigation so I left it alone.

kswenson:
@dougmartin I pushed a commit with the suggestions that were straightforward to implement and punted the ones that require actual thought.

PR #215, opened 2021-06-22
Fix loading of Google Drive API so that we only load it once

kswenson:
[[#178634302]](https://www.pivotaltracker.com/story/show/178634302)

In working through the TypeScript conversion I noticed that the GoogleDriveProvider was loading an instance of the Google Drive API in every function call rather than caching the result of the first load and returning it for all subsequent loads. This PR makes sure we only load the Google Drive API once, in consultation with @dougmartin, and also includes a jest test which in good TDD fashion was written before the fix was implemented.


dougmartin (approved):
Looks good :+1:

PR #220, opened 2021-08-04
Fix auto-configuration of InteractiveApiProvider

kswenson:
The good news is that I was able to create an example activity that demonstrates unmodified production CODAP saving/restoring files using attachments. To get it to work, however, one more tweak to provider initialization is required.


scytacki (approved):
seems good to me

PR #221, opened 2021-08-05
Working save/restore of CODAP documents as attachments

kswenson:
In setting up example activities for testing purposes I realized there was some confusion around wrapping/unwrapping attachment contents that was preventing save/restore of CODAP documents from working correctly.

The following URLs demonstrate the functionality:

CFM Demo saving with attachments:
https://activity-player.concord.org/branch/master/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F21112.json

CODAP saving without attachments (subject to Firestore 1MB document size limit):
https://activity-player.concord.org/branch/master/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F21118.json

CODAP saving with attachments:
https://activity-player.concord.org/branch/master/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F21115.json

CODAP saving with dynamic attachments (large documents (>= ~480K) saved as attachments while smaller documents are stored directly to Firestore). This hasn't really been tested but theoretically should work:
https://activity-player.concord.org/branch/master/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F21119.json

NOTE: I'm not entirely sure why the CSS string path changes were required, but I assume they're the result of some of the dependency updating I did earlier. It only affects the codap build.

### DIFF
--- a/src/code/providers/interactive-api-provider.ts
+++ b/src/code/providers/interactive-api-provider.ts
@@ -28,7 +28,14 @@ const kDynamicAttachmentSizeThreshold = 480 * 1024
 export const kAttachmentFilename = "file.json"
 
 // when writing attachments, interactive state is just a reference to the attachment
-const kInteractiveStateAttachment = { __attachment__: kAttachmentFilename }
+interface InteractiveStateAttachment {
+  __attachment__: typeof kAttachmentFilename,
+  contentType?: "application/json" | "text/plain"
+}
+const interactiveStateAttachment =
+(contentType?: InteractiveStateAttachment["contentType"]): InteractiveStateAttachment => {
+  return { __attachment__: kAttachmentFilename, contentType }
+}
 const isInteractiveStateAttachment = (content: any) =>
         (typeof content === "object") && (content.__attachment__ === kAttachmentFilename)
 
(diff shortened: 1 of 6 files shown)


src/code/providers/interactive-api-provider.ts
```diff
@@ -118,10 +125,10 @@ class InteractiveApiProvider extends ProviderInterface {
     }
   }
 
-  async readAttachmentContent() {
-    const response = await readAttachment(kAttachmentFilename)
+  async readAttachmentContent(interactiveState: InteractiveStateAttachment) {
+    const response = await readAttachment(interactiveState.__attachment__)
     if (response.ok) {
-      return response.text()
+      return interactiveState.contentType === "application/json" ? response.json() : response.text()
```
scytacki, line 131: The response should have a content type header that ought to be a safer way to determine the content type instead of looking at the interactive state.  The state could possibly get out of sync with the attachment, but the content type header should always match actual attachment file.

There might be an issue with the setup of the CORS for the S3 bucket so the browser is allowed to read the content type header in the response. 

I'm fine with punting this change so that you can move on.

scytacki (approved):
Looks great.

I think it'd be better to use the content type from the S3 response, but I think it is fine to punt that until Piotr picks this up.

PR #222, opened 2021-08-12
Fix bug in loading contents of shared CODAP documents via InteractiveApiProvider

kswenson:
This is essentially an alternative fix for @tejal-shah's bug ([[#172312303]](https://www.pivotaltracker.com/story/show/172312303)). The problem is that the state object communicated between the CFM and its client (CODAP in this case) needs be cloned so that it isn't frozen by the `lara-interactive-api` library. [LARA PR #736](https://github.com/concord-consortium/lara/pull/736) demonstrates what this would look like if we made the library responsible for cloning the state. If we decide not to do so in the library, then this PR demonstrates how it could be done in the client.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@concord-consortium/lara-interactive-api": "^1.1.2",
+    "@concord-consortium/lara-interactive-api": "1.1.2",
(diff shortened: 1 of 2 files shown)


scytacki (approved):
Looks good to me.

Now that I see this, I'm having doubts about the lara-interactive-api always freezing the object. But I still think this is the right way to go for now.

PR #226, opened 2021-09-08
[#177118798] Fix typo in CFM google provider error message

jsandoe:
* Also, add message to strings file so it can be localized.
* Updated minor version number
* Tested in CODAP: 
    * saved a file to Google Drive, 
    * then disabled wireless network, 
    * then made a change to the CODAP docuument: saw correct error message.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "cloud-file-manager",
   "description": "Wrapper for providing file management for web applications",
   "author": "The Concord Consortium",
-  "version": "1.3.0",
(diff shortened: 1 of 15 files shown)


kswenson (approved):
👍  Your changes are fine. I was going to ask you to put the string that I inadvertently put in the `en-US.json` file where it belongs in the `en-US-master.json` file, but I went ahead and did it myself to save you the trouble. You may want to re-run the push/pull scripts, however, before the PR is merged.

PR #228, opened 2021-09-17
Restore null defaults to some functions arguments that were lost in TypeScript conversion

kswenson:
In the TypeScript conversion some function arguments that had been defaulting to `null` were changed to be optional parameters instead, i.e. `undefined` would be passed instead. In most cases this doesn't matter but in some cases clients are using `===` or `!==` to test specifically for `null`. This PR restores `null` defaults to all function arguments that had them previously.


kswenson:
Merged after sync review with @jsandoe.

PR #229, opened 2021-09-21
Call cloneDeep() on initialState

kswenson:
[[#179611084]](https://www.pivotaltracker.com/story/show/179611084)

Having decided not to implement [LARA PR #736](https://github.com/concord-consortium/lara/pull/736), we instead worked around the interactive API's frozen state in [CFM PR #222](https://github.com/concord-consortium/cloud-file-manager/pull/222) by calling `cloneDeep()` in `getInteractiveState()` and `setInteractiveState()`. Unfortunately, this missed the initial interactive state, a situation that is rectified with this PR.

@scytacki I originally assigned this to Piotr for review, forgetting that he's on vacation. It should be a quick review, however, as it's essentially a trivial follow-up to #222.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "cloud-file-manager",
-  "version": "1.3.2",
+  "version": "1.3.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


scytacki (approved):
looks good to me

PR #231, opened 2021-10-15
Add support basic Linked Interactive logic  [#175171377]

dougmartin:
I was going to combine this PR with the UI story but I don't have that done so I yanked it out and just have the basic support in this PR.

To see this in action preview this activity in either Lara or AP: 

https://authoring.staging.concord.org/activities/21350

It has SageModeler using CFM master on page 1 and a linked SageModeler interactive with CFM using this branch on page 2.  You should be able to update the text box in the SageModeler area on page 1 and then see the update on page 2.

### DIFF
--- a/src/code/providers/interactive-api-provider.ts
+++ b/src/code/providers/interactive-api-provider.ts
@@ -146,12 +146,20 @@ class InteractiveApiProvider extends ProviderInterface {
             : cloneDeep(interactiveState)
   }
 
+  getInitialInteractiveState(initInteractiveMessage: IRuntimeInitInteractive) {
+    let interactiveState = initInteractiveMessage.interactiveState
+
+    if (!interactiveState && (initInteractiveMessage.hasLinkedInteractive && initInteractiveMessage.linkedState)) {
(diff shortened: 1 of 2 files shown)


src/code/providers/interactive-api-provider.ts
```diff
@@ -148,8 +148,13 @@ class InteractiveApiProvider extends ProviderInterface {
...
+    const initialInteractiveState = initInteractiveMessage.hasLinkedInteractive && initInteractiveMessage.linkedState
+      ? initInteractiveMessage.linkedState
+      : initInteractiveMessage.interactiveState
+
     try {
-      interactiveState = await this.processRawInteractiveState(initInteractiveMessage.interactiveState)
+      interactiveState = await this.processRawInteractiveState(initialInteractiveState)
     }
     catch(e) {
       // on initial interactive state there's not much we can do on error besides ignore it
```
dougmartin, line 160: I thought about falling back to the interactiveState here if the linkedState has an error but think it would be better to surface that as an error.  I plan on adding that as part of the UI work.

scytacki (approved):
looks good to me

PR #232, opened 2021-10-19
Support Edge case Linked Interactive logic and UI [#179625030]

dougmartin:
Remaining work:

- Uncomment "touch" setInteractiveState with AP supports it (follow on PR)

### DIFF
--- a/src/code/providers/interactive-api-provider.ts
+++ b/src/code/providers/interactive-api-provider.ts
@@ -200,10 +200,12 @@ class InteractiveApiProvider extends ProviderInterface {
           interactiveStateAvailable
         })
 
-        // remove existing interactive state, so the interactive will be initialized from the linked state next time (if it is not saved).
-        if (interactiveState === mostRecentLinkedState) {
-          // TODO: figure out why this isn't working (and doesn't work with setTimeout(...) either)
+        if (interactiveState === mostRecentLinkedState.interactiveState) {
+          // remove existing interactive state, so the interactive will be initialized from the linked state next time (if it is not saved).
           setInteractiveState(null)
+        } else {
+          // TODO: add support in AP for "touch" to update
+          // setInteractiveState("touch")
         }
 
         return interactiveState
--- a/src/code/providers/interactive-api-provider.ts
+++ b/src/code/providers/interactive-api-provider.ts
@@ -134,8 +134,8 @@ class InteractiveApiProvider extends ProviderInterface {
     }
   }
 
-  async readAttachmentContent(interactiveState: InteractiveStateAttachment) {
-    const response = await readAttachment(interactiveState.__attachment__)
+  async readAttachmentContent(interactiveState: InteractiveStateAttachment, questionId?: string) {
+    const response = await readAttachment({name: interactiveState.__attachment__, questionId})
     if (response.ok) {
       // TODO: Scott suggests reading contentType from response rather than from interactiveState
       return interactiveState.contentType === "application/json" ? response.json() : response.text()
@@ -145,9 +145,9 @@ class InteractiveApiProvider extends ProviderInterface {
     }
   }
 
-  async processRawInteractiveState(interactiveState: any) {
+  async processRawInteractiveState(interactiveState: any, questionId?: string) {
     return isInteractiveStateAttachment(interactiveState)
-            ? await this.readAttachmentContent(interactiveState)
+            ? await this.readAttachmentContent(interactiveState, questionId)
             : cloneDeep(interactiveState)
   }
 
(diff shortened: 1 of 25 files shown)


dougmartin:
> * the core CFM code has to change in order for a provider to show a custom dialog. I don't think this needs to be addressed because we don't add lots of providers.

Yes, this is the only case were we need a special dialog for the provider but given the design it can't be helped.

> * the adding of english strings to the other lang files seems strange, I'd hope that if there is no string in another language that it would just fall back to english so these changes shouldn't be necessary. But I'm guessing that isn't how it works.

That is how POEditor works for these projects.  We update the local en-us.master.json file and push it to POEditor and then immediately pull the strings.  POEditor puts the new keys in all of the languages defined for the project and defaults to the english version so when we pull the other languages have English as the default until they are translated.

> * I don't see how the page number, page name, and activity name get to the dialog. Are all of these included in the list of interactive states provided by the interactive host? If so the mock state in the test should be updated.

The page & activity info is in the outer initInteractiveMessage and in each of the allLinkedStates objects.  Those tests are really just checking the interactive state picked so they don't really exercise the dialog at all.

src/code/providers/interactive-api-provider.ts
```diff
@@ -146,20 +151,89 @@ class InteractiveApiProvider extends ProviderInterface {
...
+      } else {
+        // currently the AP doesn't make the updatedAt attribute available so just pick the directly linked state
+        mostRecentLinkedState = directlyLinkedState
+      }
+
+      const currentDataTimestamp = new Date(initInteractiveMessage.updatedAt || Date.now())
+      const mostRecentLinkedStateTimestamp = new Date(mostRecentLinkedState.updatedAt || Date.now())
+      const directlyLinkedStateTimestamp = new Date(directlyLinkedState.updatedAt ||  Date.now())
+
+      // current state is available, but there's most recent data in one of the linked states. Ask user.
```
kswenson, line 195:
```suggestion
      // current state is available, but there's more recent data in one of the linked states. Ask user.
```

src/code/providers/interactive-api-provider.ts
```diff
@@ -146,20 +151,89 @@ class InteractiveApiProvider extends ProviderInterface {
...
+
     let interactiveState = initInteractiveMessage.interactiveState
 
-    if (!interactiveState && (initInteractiveMessage.hasLinkedInteractive && initInteractiveMessage.linkedState)) {
-      interactiveState = initInteractiveMessage.linkedState
+    const interactiveStateAvailable = !!interactiveState
+    const {allLinkedStates} = initInteractiveMessage
+
+    // this is adapted from the existing autolaunch.ts file
+    if (allLinkedStates && allLinkedStates.length > 0) {
```
kswenson, line 177: Could also use
```suggestion
    if (allLinkedStates?.length) {
```

src/code/utils/lang/de.json
```diff
@@ -125,5 +125,14 @@
...
+    "~SHARE_DIALOG.INTERACTIVE_API_MESSAGE": "Use this link when creating an activity for the Activity Player",
+    "~DIALOG.SELECT_INTERACTIVE_STATE.USE_THIS_VERSION": "Use this version",
+    "~DIALOG.SELECT_INTERACTIVE_STATE.CLICK_TO_PREVIEW": "Click to preview",
+    "~DIALOG.SELECT_INTERACTIVE_STATE.UPDATED_AT": "Updated at",
+    "~DIALOG.SELECT_INTERACTIVE_STATE.PAGE": "Page",
+    "~DIALOG.SELECT_INTERACTIVE_STATE.ACTIVITY": "Activity",
+    "~DIALOG.SELECT_INTERACTIVE_STATE.CURRENT_VS_LINKED": "Another page contains more recent data. Which would you like to use?",
+    "~DIALOG.SELECT_INTERACTIVE_STATE.LINKED_VS_LINKED": "There are two possibilities for continuing your work. Which version would you like to use?",
+    "~DIALOG.SELECT_INTERACTIVE_STATE.TITLE": "What would you like to do?",
+    "~DIALOG.SELECT_INTERACTIVE_STATE.PREVIEW_INFO": "This is read-only preview of your data. Click anywhere to close it."
```
kswenson, line 137:
```suggestion
    "~DIALOG.SELECT_INTERACTIVE_STATE.PREVIEW_INFO": "This is a read-only preview of your data. Click anywhere to close it."
```

src/code/views/select-interactive-state-dialog-view.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+interface SelectInteractiveStateVersionProps {
+  version: IInteractiveStateProps<{}>
+  showingOverlay: boolean
+  onSelect?: SelectInteractiveStateCallback
+  onPreview: (showing: boolean) => void
+  close?: () => void
+}
+
+interface SelectInteractiveStateVersionState {
```
kswenson, line 15: Since `SelectInteractiveStateVersionProps` and `SelectInteractiveStateVersionState` are only used internally, you could use `IProps` and `IState` which would improve readability below.

src/code/views/select-interactive-state-dialog-view.tsx
```diff
@@ -0,0 +1,160 @@
...
+  showOverlay: boolean
+}
+
+export default class SelectInteractiveStateDialog extends React.Component<SelectInteractiveStateDialogProps, SelectInteractiveStateDialogState> {
+
+  constructor(props: SelectInteractiveStateDialogProps) {
+    super(props)
+
+    this.handleOnPreview = this.handleOnPreview.bind(this)
+    this.handleHideOverlay = this.handleHideOverlay.bind(this)
```
kswenson, line 117: Given that these methods are defined as function expressions I don't think the binding should be necessary here.

kswenson (approved):
👍 Looks good -- just a few minor comments/suggestions

PR #233, opened 2021-11-10
179981050 handle gd auth timeout

jsandoe:
- Relevant story here: https://www.pivotaltracker.com/story/show/179981050
- Version number has not yet updated
- The changes affect providers that authenticate, which I think are only GD and the Document Server provider. I did not test the later, but I believe it is not in use any more. I think, for example, that document sharing is provided through another provider.
- These changes were built off of the then head of the `master` branch. This includes code from PR #231, which have not been included in any CODAP build. Although it is on `master`, I am not sure of its readiness for production nor PR #232 which was merged subsequently. If these are ready for release, I can rebase as appropriate.

### DIFF
--- a/src/code/client.ts
+++ b/src/code/client.ts
@@ -497,20 +497,28 @@ class CloudFileManagerClient {
     }
   }
 
+  closeCurrentFile() {
+    console.log('Closing file (rejected reauth)');
+    this.closeFile();
+  }
+
   confirmAuthorizeAndOpen(provider: ProviderInterface, providerParams: any) {
+    let rejectCallback = function () {this.closeCurrentFile();}.bind(this);
     // trigger authorize() from confirmation dialog to avoid popup blockers
     return this.confirm(tr("~CONFIRM.AUTHORIZE_OPEN"), () => {
-      return provider.authorize(() => {
-        this._event('willOpenFile', {op: "confirmAuthorizeAndOpen"})
-        return provider.openSaved(providerParams, (err: string | null, content: any, metadata: CloudMetadata) => {
-          if (err) {
-            return this.alert(err)
-          }
-          this._fileOpened(content, metadata, {openedContent: content.clone()}, this._getHashParams(metadata))
-          return provider.fileOpened(content, metadata)
+        return provider.authorize(() => {
+          this._event('willOpenFile', {op: "confirmAuthorizeAndOpen"})
+          return provider.openSaved(providerParams, (err: string | null, content: any, metadata: CloudMetadata) => {
+            if (err) {
+              return this.alert(err)
+            }
+            this._fileOpened(content, metadata, {openedContent: content.clone()}, this._getHashParams(metadata))
+            return provider.fileOpened(content, metadata)
+          })
         })
-      })
-    })
+      },
+      rejectCallback
+    );
   }
 
   openProviderFileWhenConnected(providerName: string, providerParams?: any) {
@@ -584,12 +592,14 @@ class CloudFileManagerClient {
   }
 
   confirmAuthorizeAndSave(stringContent: any, callback?: OpenSaveCallback) {
+    let rejectCallback = function () {this.closeCurrentFile();}.bind(this);
     // trigger authorize() from confirmation dialog to avoid popup blockers
     return this.confirm(tr("~CONFIRM.AUTHORIZE_SAVE"), () => {
       return this.state.metadata.provider.authorize(() => {
         return this.saveFile(stringContent, this.state.metadata, callback)
       })
-    })
+    },
+    rejectCallback)
   }
 
   save(callback: OpenSaveCallback = null) {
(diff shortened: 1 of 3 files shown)


dougmartin:
@jsandoe I just merged https://github.com/concord-consortium/cloud-file-manager/pull/232 - could you rebase and then I'll review?

jsandoe:
Sure @dougmartin , I rebased my changes on top of #232 and force-pushed.

src/code/client.ts
```diff
@@ -1049,7 +1069,7 @@ class CloudFileManagerClient {
       interval = Math.round(interval / 1000)
     }
     if (interval > 0) {
-      return this._autoSaveInterval = window.setInterval((() => { if (this.shouldAutoSave()) { return this.save() } }), interval * 1000)
+      return this._autoSaveInterval = window.setInterval((() => { if (this.shouldAutoSave()) { console.log('autosaving'); return this.save() } }), interval * 1000)
```
dougmartin, line 1072: This new `console.log` is going to add a lot of noise to the console as it will print each second.

dougmartin (approved):
Looks good :+1:

src/code/client.ts
```diff
@@ -645,6 +663,8 @@ class CloudFileManagerClient {
             return this.alert(err)
           }
         }
+        metadata.autoSaveDisabled = true;
+        this._setState({ metadata, saving: null })
```
kswenson, line 667: I think this warrants a comment. The effect seems to be that we disable autosave after the second non-40x failure but not the first, but I'm not sure I understand the rationale.

src/code/client.ts
kswenson: I presume this can be removed.

src/code/providers/google-drive-provider.ts
```diff
@@ -176,6 +189,11 @@ class GoogleDriveProvider extends ProviderInterface {
     })
   }
 
+  authorize(callback:any) {
+    this.authCallback = callback;
+    this.doAuthorize(!GoogleDriveProvider.IMMEDIATE);
```
kswenson, line 194: I think a comment would be useful here as well.

src/code/providers/google-drive-provider.ts
```diff
@@ -278,7 +307,9 @@ class GoogleDriveProvider extends ProviderInterface {
     })
   }
 
-  close(metadata: CloudMetadata, callback: ProviderCloseCallback) {}
+  close(metadata: CloudMetadata, callback: ProviderCloseCallback) {
+    console.log('Closing GD File');
```
kswenson, line 311: Seems like this could get noisy as well. Was this for debugging or were you proposing keeping it moving forward?

src/code/client.ts
```diff
@@ -497,20 +497,37 @@ class CloudFileManagerClient {
...
   }
 
+  /**
+   * Disassociates the current document from its provider.
+   *
+   * This is important specifically for autosaving providers. When authenticated
+   * state is lost we must not continue to autosave.
+   */
+  disconnectCurrentFile() {
+    console.log('Closing file (rejected reauth)');
```
kswenson, line 507:
```suggestion
    console.warn('Closing file (rejected reauth)');
```
My convention is to use `console.warn()` for events worth noting in the logs on production and `console.log()` for events used locally when debugging but that should be removed before releasing. This seems like an event worth noting in the production logs.

kswenson (commented):
[Chiming in unbidden]
👍 Looks good! It'll be nice to improve this behavior. I made several minor suggestions for linting, minor code tweaks, and adding some comments.

jsandoe:
Thank you for the review and the suggestions, @dougmartin and @kswenson. I will implement the suggestions and resubmit. Are the changes for the Interactive provider that have been merged into master at a point that they can be included in a CODAP release?

dougmartin:
@jsandoe the changes to the CFM for the interactive provider are being QAed now.  If you can wait a couple of days I think we will be good to go.

src/code/providers/google-drive-provider.ts
jsandoe: It was for debugging. Removing.

src/code/client.ts
jsandoe: Thanks for spotting this! That was definitely not my intention. Clearly this change was poorly thought out. Will move these lines to where they will have better effect.

jsandoe:
@kswenson , I have made the fixes responding to your comments. Ready for your recheck.

src/code/client.ts
```diff
@@ -584,12 +601,14 @@ class CloudFileManagerClient {
   }
 
   confirmAuthorizeAndSave(stringContent: any, callback?: OpenSaveCallback) {
+    let rejectCallback = function () {this.disconnectCurrentFile();}.bind(this);
```
kswenson, line 604:
```suggestion
    const rejectCallback = () => this.disconnectCurrentFile()
```

kswenson (approved):
👍 Looks good! Just one lingering semicolon/bind that didn't get removed.

kswenson:
>Damn! I must have a block about this!

😆 I would have to agree, as you removed one semicolon but left another as well as the unnecessary `bind`. My apologies if my imprecise "Just one lingering semicolon/bind" comment led you astray.

PR #234, opened 2021-11-17
Fixed scaling on linked interactive previews so that it is readable

dougmartin:
This fixes a long standing feature request from Dan - when the smaller linked interactive preview is clicked it is displayed in a larger area but the scaling stayed the same so it was very hard to read.  This also presented a problem during QA so I went ahead and fixed it (and it turned out it was a pretty easy fix)

Existing (unchanged preview selector):
image

Scaling **BEFORE** Fix
image

Scaling **AFTER** Fix (this PR)
image


knowuh (approved):
Looks good

PR #236, opened 2021-11-24
Send `requiresUserInteraction` event when putting up state selection dialog

kswenson:
[[#180401986]](https://www.pivotaltracker.com/story/show/180401986)

CODAP presents the splash screen but makes decisions about when to show/hide it based on event notifications from CFM. In this particular instance, CFM needs to show a dialog itself with which the CODAP splash screen interferes. Therefore, CFM sends the new `requiresUserInteraction` message when it needs to present a dialog during startup and in a separate [CODAP PR](https://github.com/concord-consortium/codap/pull/398), CODAP responds by hiding the splash screen when the CFM requires user interaction.

Also, increment the version number, fix a typo, and run `npm audit fix` while we're in the neighborhood.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "version": "1.4.1",
   "repository": {
     "type": "git",
-    "url": "https://github.com/concord-consortium/cloud-file-manger"
+    "url": "https://github.com/concord-consortium/cloud-file-manager"
   },
   "license": "MIT",
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "cloud-file-manager",
   "description": "Wrapper for providing file management for web applications",
   "author": "The Concord Consortium",
-  "version": "1.4.0",
+  "version": "1.4.1",
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "cloud-file-manager",
   "description": "Wrapper for providing file management for web applications",
   "author": "The Concord Consortium",
-  "version": "1.4.0",
+  "version": "1.4.1",
   "repository": {
     "type": "git",
     "url": "https://github.com/concord-consortium/cloud-file-manger"
```
dougmartin, line 8: Not needed for this PR but I just noticed the repository url has a typo:

`"url": "https://github.com/concord-consortium/cloud-file-manger"`

should be

`"url": "https://github.com/concord-consortium/cloud-file-manager"`

dougmartin (approved):
Looks good :+1:

jsandoe (approved):
Looks good.

package.json
kswenson: I added a commit to fix the typo as long as I'm in the neighborhood.

PR #237, opened 2021-12-16
Added save hash check and fixed incorrect saved flag setting on file loads [#180438651]

dougmartin:
This adds a early exit on saves if the hash of the content doesn't change between the first load and save or between commanded saves.

This also adds a test provider so that we can use jest tests to test the new code that was added.

The "saved" flag was being set on the load event instead of the dirty flag.  The update is to set the dirty flag to true if the content requires conversion.

### DIFF
--- a/src/code/client.ts
+++ b/src/code/client.ts
@@ -7,6 +7,8 @@
  * Full docs: https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md
  */
 import _ from 'lodash'
+import { sha256 } from 'js-sha256'
+
 import tr from './utils/translate'
 import isString from './utils/is-string'
 import base64Array from 'base64-js' // https://github.com/beatgammit/base64-js
@@ -26,6 +28,7 @@ import S3Provider  from './providers/s3-provider'
 import LocalFileProvider  from './providers/local-file-provider'
 import PostMessageProvider  from './providers/post-message-provider'
 import URLProvider  from './providers/url-provider'
+import TestProvider  from './providers/test-provider'
 
 import {
   CloudContent, cloudContentFactory, CloudMetadata, ECapabilities, ProviderInterface
@@ -47,6 +50,7 @@ interface IClientState {
   sharing?: boolean;
   dirty?: boolean;
   failures?: number;
+  contentHash?: string;
 }
 
 export type ClientEventCallback = (...args: any) => void;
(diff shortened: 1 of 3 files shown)


src/code/client.ts
```diff
@@ -545,8 +550,9 @@ class CloudFileManagerClient {
             if (err) {
               return this.alert(err, () => this.ready())
             }
-            // if we just opened the file, it doesn't need to be saved until the contents are changed
-            const additionalState = { openedContent: content.clone(), saved: true }
+            // if we just opened the file, it doesn't need to be saved until the contents are changed unless
+            // it requires conversion from an older version
+            const additionalState = { openedContent: content.clone(), dirty: content.requiresConversion() }
```
kswenson, line 555: >If you are around in the next hour or so, I would love to just quickly get your feedback on small CFM change that Doug and I just made.

>The `saved: true` seemed to be triggering a UI update saying that *all changes have been saved* which was confusing us in QA because we were thinking that a save should not have been triggered.
>We also set the dirty bit check if the loaded content needs to be updated. A similar thing is done elsewhere in the client when the content updates.

Seems right to me. You never know if there might be other downstream code that is relying on this behavior somehow, but I bet whoever added that `saved: true` was thinking of it as a dirty flag. In any case, this seems like the correct behavior of this function and we can deal with any downstream consequences that might arise.

src/code/client.ts
```diff
@@ -634,6 +640,15 @@ class CloudFileManagerClient {
   }
 
   saveFile(stringContent: any, metadata: CloudMetadata, callback: OpenSaveCallback = null) {
+    // if the content didn't change skip the save (but fake it for the client)
+    const savedContentHash = this._computeContentHash(stringContent)
+    if (this.state.contentHash === savedContentHash) {
+      console.log("CFM: File content not changed, skipping sending save to provider!")
```
kswenson, line 646: Were you planning to remove this once we're sure it's working correctly?

kswenson (approved):
👍 LGTM -- just one question which you don't have to do anything about. 😉

src/code/client.ts
```diff
@@ -7,6 +7,8 @@
  * Full docs: https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md
  */
 import _ from 'lodash'
+import { sha256 } from 'js-sha256'
```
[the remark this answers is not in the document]
dougmartin: It seemed fast enough during testing and its already a dependency so that is why I went with it.

src/code/client.ts
dougmartin: No, I want to leave it in the code in case we get any reports of strange behavior - it will be easier to identify that the save was faked.

PR #241, opened 2022-01-10
Lang pt br

jsandoe:
Add translation for Brazilian Portuguese.


src/code/utils/translate.ts
```diff
@@ -36,6 +37,7 @@ const languageFiles: LanguageFileEntry[] = [
   {key: 'ja' ,   contents: ja},     // Japanese
   {key: 'nb',    contents: nb},     // Norwegian Bokmål
   {key: 'nn',    contents: nn},     // Norwegian Nynorsk
+  {key: 'pt-BR',    contents: ptBR},     // Brazilian Portuguese
```
kswenson, line 40:
```suggestion
  {key: 'nn',    contents: nn},     // Norwegian Nynorsk
  {key: 'pt-BR', contents: ptBR},   // Brazilian Portuguese
```

kswenson (changes requested):
👍 Looks good -- just a couple of suggested tweaks
- adjust white space in language list to align with surrounding code
- now that I better understand the context of your prior slack questions, yes, I think you should also bump the version number given that you're making a CFM change. You could do that as a separate PR but I think it'd also be fine to bump the version number as part of this PR since we've essentially already decided that this is the last change for this version. As for what the new version should be, is adding a translation a feature (in which case it should be 1.6.0) or not (in which case it should be 1.5.2)? I'm leaning towards 1.5.2, but I'm not sure I have a compelling argument for that gut feeling.

jsandoe:
Agree with comments. Updated PR to correct formatting and bump the minor version number.

kswenson (approved):
👍 Looks good -- in theory, the version number should be changed in `package-lock.json` also (e.g. by running `npm install`) but I don't know that it matters.

jsandoe:
Good thought. Updated package-lock.json.

PR #242, opened 2022-01-12
Fixed relative path of earth.svg in language menu

dougmartin:

PR #243, opened 2022-01-12
[#180892247] Fix bug: icon font path broken for CODAP Build.

jsandoe:
There was a recent adjustment to font paths for the Codap-Ivy icon font. Apparently this was to fix an error in the Sage build. It broke the references in the CODAP build. As far as I know, the only symbols actually used in the CODAP build are for file list icons in the Open/Save/... dialogs. Fixed the scripts for the CODAP build to restore the original paths.

### DIFF
--- a/build-support/build-opts.js
+++ b/build-support/build-opts.js
@@ -34,7 +34,11 @@ const replacementStrings = {
 if (codap) {
   replacementStrings.css.push(
     {
-      search: /url\(..\/..\//g,
+      search: /url\("\.\.\//g,
+      replace: 'url("'
+    },
+    {
+      search: /url\(/g,
       replace: 'static_url('
     },
     {
--- a/build-support/build-opts.js
+++ b/build-support/build-opts.js
@@ -34,7 +34,7 @@ const replacementStrings = {
 if (codap) {
   replacementStrings.css.push(
     {
-      search: /url\(/g,
+      search: /url\(..\/..\//g,
       replace: 'static_url('
     },
     {


build-support/build-opts.js
```diff
@@ -34,7 +34,7 @@ const replacementStrings = {
 if (codap) {
   replacementStrings.css.push(
     {
-      search: /url\(/g,
+      search: /url\(..\/..\//g,
```
kswenson, line 37: I may be missing something here, but
1. Wouldn't those `.`s match any character? It seems like you probably want something like: `/url\(\.\.\/\.\.\//g`, which admittedly looks pretty ridiculous, but 🤷 .
2. Wouldn't this break the other images and fonts that are at `../` paths?
It seems like we need something that catches the problematic font but doesn't change the behavior for everything else. What was the original change that required this change?

kswenson (changes requested):
I think I need more context to make sense of this.

build-support/build-opts.js
jsandoe: I agree with (1). Dumb error on my part.
Concerning (2), there aren't others with the "../../" pattern in the styles directory. It was serendipity that the original RE worked at all. It altered other urls, but these are unused in CODAP. Some other process in the build script apparently strips one of the "../" substrings from urls. 

Have uploaded a better fix. It only alters the broken font URLs.

kswenson (approved):
👍 Looks good -- I was confused initially because when looking at the _source_ files (e.g. the `.styl` files),  the `CodapIvy` files are at paths like `url('../../fonts/CodapIvy.eot')` and there are a number of other assets that are at paths like `url('../some/asset/path')`, so it seemed like your change could adversely affect other assets. But, of course, the string replacement is occurring on the _output_ files, where the `CodapIvy` files have paths like `url("../fonts/CodapIvy.eot")` and the other asset files do not, which explains why your change only affects the `CodapIvy` files, as was your original intent.

dougmartin:
NOTE: Dan also noticed the Earth icon in the language switcher menu was broken so I have an unmerged PR to fix that:

https://github.com/concord-consortium/cloud-file-manager/pull/242/files

I'm not sure if this effects CODAP but I just wanted to point that out.

PR #245, opened 2022-01-20
Fix loss of file extension on rename after copy [#174494057]

dougmartin:
This changes the code to use the newish rename method on the CloudMetadata class which handles setting the filename extension automatically.


kswenson (approved):
👍 Looks good!

PR #246, opened 2022-01-24
Save query string on copy [#174494057]

dougmartin:
This fix is really for testing so that alternate cfm branches spec'ed in the query string are included in the copied tab.


kswenson (approved):
👍 Looks good -- I was going to point out that this would break any clients that were relying on the previous query-string handling behavior of the function but a quick check of the code indicates that all clients are in fact passing hash strings, so your changes are just aligning the implementation with the actual usage.

PR #250, opened 2022-01-25
Fix autosave on language change [#175390295]

dougmartin:
Before this fix the autosave was short-circuited if the document wasn't dirty, now it always saves.

This fixes a bug where if the user changed the language more than once the original document wasn't saved the second time because it wasn't dirty after the first language change save.


knowuh (approved):
Looks good, thanks

PR #251, opened 2022-01-27
Remove autosave of readonly providers in language switching [#175390295]

dougmartin:
Before this change when an example file was loaded and then the language was changed a save dialog would be shown as the readonly provider can't save but there were other providers that could, eg Google Drive.

With this change the current provider saves only if it can, otherwise it falls back to saving a temp file in local storage.


knowuh (approved):
LGTM

PR #254, opened 2022-02-03
Ensure document id in interactive api is not url encoded [#181103123]

dougmartin:
Some of the question interactives have encoded CODAP/SageModeler urls that themselves have encoded document id urls.  This fix ensures the document id urls are fully decoded before they are fetched.


scytacki (changes requested):
I think it isn't a good idea to do this.
Whatever is launching the CFM should decode the params correctly instead.

Perhaps the documentId will have some characters in it that shouldn't be decoded twice.

dougmartin:
Deleting.

PR #257, opened 2022-02-22
Build CODAP version in production mode by default

jsandoe:
Current size of the CFM: 2M. Production mode size: 800k.


kswenson (approved):
👍 LGTM -- I'm surprised it makes that much of a difference. Good catch!

PR #259, opened 2022-03-07
Change to default to dynamic attachments with length check [#181405875]

dougmartin:


src/code/providers/interactive-api-provider.ts
```diff
@@ -85,12 +83,15 @@ class InteractiveApiProvider extends ProviderInterface {
...
-        return JSON.stringify(content).length >= kDynamicAttachmentSizeThreshold
+    if (interactiveApi === kAttachmentUrlParameter) {
+      return true
     }
+
+    const aboveDynamicThreshold = JSON.stringify(content).length >= kDynamicAttachmentSizeThreshold
+    if (aboveDynamicThreshold) {
+      return true
+    }
+
```
kswenson, line 94:
```suggestion
     return aboveDynamicThreshold
```

kswenson (approved):
👍 Looks good -- just one small code simplification suggestion

scytacki (approved):
Seems good to me.

I was hoping you'd find a bug in here that was causing it to not work when the size started small and grew too large. But I can see that you didn't. And I looked at the code and don't see how that kind of bug could happen.

PR #261, opened 2022-03-09
Change interactive member of init message optional [#181103123]

dougmartin:
Some interactives, like the full-screen wrapper always report they are in runtime  mode, even when loaded in a report which does not define the interactive member.  This change makes the interactive member of the message optional.


scytacki (approved):
Seems fine to me.

Adding to the comment that this means linked states will not work when this interactiveId is not sent could be useful, but I guess it is pretty obvious from the code.

PR #275, opened 2022-08-19
Added rewrite of sensor-interactive domain in interactive state links [#182916725]

dougmartin:
Also added `iframeAllow` client option so that SageModeler (and other apps that use CFMs wrapper) to inject `bluetooth` and other allowed features into the outer app iframe.


scytacki (approved):
looks good to me

PR #279, opened 2022-11-02
Moved attachment creation to inside setInteractiveState [#183681726]

dougmartin:
This fixes a bug, surfaced via the StoryQ CODAP plugin, where a large initial interactive state in a shared document was loaded and then saved on startup, bypassing the attachment checking in the save method.

This also fixes an error, found during this fix, of the attachment state being leaked out of the save call via the callback.  It should instead return the actual saved state in that case, not the attachment wrapper.

### DIFF
--- a/src/code/providers/interactive-api-provider.ts
+++ b/src/code/providers/interactive-api-provider.ts
@@ -13,12 +13,43 @@ import {
 } from '@concord-consortium/lara-interactive-api'
 import { SelectInteractiveStateDialogProps } from '../views/select-interactive-state-dialog-view'
 
+export const shouldSaveAsAttachment = (content: any) => {
+  const interactiveApi = queryString.parse(location.search).interactiveApi
+  if (interactiveApi === kAttachmentUrlParameter) {
+    return true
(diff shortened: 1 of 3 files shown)


src/code/providers/interactive-api-provider.ts
```diff
@@ -13,12 +13,43 @@ import {
...
+  if (interactiveApi === kAttachmentUrlParameter) {
+    return true
+  }
+
+  const aboveDynamicThreshold = JSON.stringify(content).length >= kDynamicAttachmentSizeThreshold
+  if (aboveDynamicThreshold) {
+    return true
+  }
+
+  return false
```
kswenson, line 27: Could just:
```
  return aboveDynamicThreshold
```

kswenson (approved):
👍 Looks good

src/code/providers/interactive-api-provider.ts
dougmartin: Yep - I just copied that function without change out of the class so it could be used outside of it.

PR #282, opened 2022-11-29
Added autosaving of directly linked state when there is no current state [#183903716]

dougmartin:
Before this change if a student shared a Sage interactive which was displaying linked state instead of that question's interactive state (as the student had not made changes yet) the sharing plugin would show an empty document instead of the linked state seen by the student.  The sharing plugin uses the portal-report which uses the report initInteractive message which does not send linked interactive states as part of its message (only the runtime message does).  By autosaving the directly linked state when there is no current state the sharing plugin will now mirror what the student sees.

### DIFF
--- a/src/code/providers/interactive-api-provider.test.ts
+++ b/src/code/providers/interactive-api-provider.test.ts
@@ -335,7 +335,9 @@ describe('InteractiveApiProvider', () => {
     expect(getMockedCall(clientListener, "openedFile").data.content).toEqual({foo: "test"})
     expect(mockFetch).not.toHaveBeenCalled()
     expect(mockApi.getInteractiveState).not.toHaveBeenCalled()
-    expect(mockApi.setInteractiveState).not.toHaveBeenCalled()
+
+    // check that the linked state is autosaved
+    expect(mockApi.setInteractiveState).toHaveBeenCalled()
   })
 
   it("uses interactive state if interactive state and linked state is provided", async () => {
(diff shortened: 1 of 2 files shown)


pjanik (commented):
If I understand correctly, this would make the interactive save itself automatically on the page load when it's linked to another interactive? Wouldn't that mess up the reports? 

I think we usually avoid autosaving interactive on page load, as then the report will show question/interactve as answered, even if the user didn't interact with it at all. 

Isn't that the same case? 

In this case, maybe we should approach it from the other side, eg:

> The sharing plugin uses the portal-report which uses the report initInteractive message which does not send linked interactive states as part of its message (only the runtime message does)

send linked interactive states as part of report initInteractive message.

But it's just my speculation, as I'm pretty out of context.

dougmartin:
Good question.  The report case is handed at the top of the function with this code:

```
    if (initInteractiveMessage.mode === "report") {
      return {interactiveState: initInteractiveMessage.interactiveState}
    }
```

The report `initInteractive` message doesn't contain any of the linked state information, just the current interactive state, which is why it is already handled at the top - the type checking would fail otherwise.

pjanik (approved):
I discussed that more with @dougmartin on slack.
It seems like it's a good fix for now. 

Long-term, I would consider if we shouldn't go this way:
> I thought the best solution would be to update the portal-report and lara interactive api to pass the linked interactive info so it could return the linked state.

But this might not be worth the effort if Dan is fine with the current behavior and he's the only user of this interactive.

PR #286, opened 2022-12-01
Treat empty object as empty content

kswenson:
PT: [[#183923855]](https://www.pivotaltracker.com/story/show/183923855)

Under some circumstances that we don't entirely understand, but seemingly involving a failure to save user content, the interactive state ends up being an empty object (`{}`), which is considered invalid by CODAP. Until we fix the underlying bug, we treat an empty object as null/empty content, which allows the authored content to be shown in the situation described by the PT story.


dougmartin (approved):
Looks good :+1:

PR #293, opened 2022-12-16
Update the new Google auth flow [#181522476]

dougmartin:


pjanik (approved):
I have zero experience with this part of CFM and Google auth flow, so I might not be the best person to review. But I guess it might be the same case for @lublagg. But things look good. If there are more changes upcoming, maybe there's someone else who has more context (Kirk?).

PR #295, opened 2022-12-21
Add loop to Google file listing to get all files [#166448174]

dougmartin:
The file api call returns a token if there are more results than can fit in the returned results.  This change keeps looping until there are no more results.


lublagg (approved):
Looks good to me 👍

pjanik (approved):
Looks good 👍

PR #296, opened 2022-12-22
Fixed renaming with Google Drive provider [#183722341]

dougmartin:
Between v2 and v3 the client method changed from patch to update and the parameter changed from title to name.


lublagg (approved):
Looks good 👍

PR #298, opened 2023-01-04
Fix issue with CODAP file dialog buttons [#184137566]

dougmartin:
This adds file dialog specific button styling to fix an issue with CODAP rendering the buttons outside the dialog due to its custom global navbar button css.

The issue is caused by the new filterable file listings.  As the height changes on the file list the buttons were not adjusting.


dougmartin:
Turns out this isn't needed - the issue was the new dialogClearFilter class needs to be added to CODAP, which it will in the next CFM import.

PR #299, opened 2023-01-05
Added autosave after rename [#184152448]

dougmartin:
Before the saved file was not autosaved after renaming.  This adds the autosave if a provider is available which then shows save feedback in the nav.


lublagg (approved):
Looks good 👍

PR #304, opened 2023-01-10
Removed LARA tab from sharing dialog [#184184524]

dougmartin:
Also fixed long standing css issues with the share loading dialog:

1. Set svg loading spinner size to 16px x 16px
2. Centered the spinner and text both horizontally and vertically in dialog


kswenson (approved):
👍 LGTM

PR #307, opened 2023-01-11
Added forced authorization when opening Google Drive files from url param [#184053851]

dougmartin:
The new Google auth flow requires a popup to show for authorization before auto opening a file using the url parameters.


lublagg (approved):
Looks good 👍

PR #308, opened 2023-01-11
Added check of extension of saves [#143693677]

dougmartin:
This allows the user to export a png or svg from CODAP with the same name as the CODAP document.


lublagg (approved):
Looks good to me 👍

PR #309, opened 2023-01-17
Fixed issues with new file filter dialog on saves [#183367471]

dougmartin:
- On save, don't clear the search filter when a folder is selected
- On save, update the search filter when a file is selected
- Fix file lookup when confirming save over existing file


lublagg (approved):
Looks good to me!

PR #310, opened 2023-01-17
Fixed blank parent folder when showing open dialog the second time [#184192561]

dougmartin:
The logic should have been: is this the initial folder or has the provider changed and instead it was is this the initial folder and has the provider changed.


lublagg (approved):
Looks good 👍

PR #311, opened 2023-01-18
Further fix for changed provider when getting folder state [#184192561]

dougmartin:


dougmartin:
Quick change after Dan testing - merging without review.

PR #314, opened 2023-01-23
Add alert when copy fails [#184239674]

dougmartin:
Shows an alert to the user with helpful next steps if the copy fails.

image


lublagg (approved):
Looks good :)

PR #317, opened 2023-01-30
Added better read-only support [#180756824]

dougmartin:
Instead of showing a save error on readonly shared Google drive files the save dialog is always shown on saves.


lublagg (approved):
Looks good 👍

PR #318, opened 2023-01-30
Updated share dialog text [#184351268]

dougmartin:
Also modernized styling to replace float with flex layout.


dougmartin:
Merging without review as it is only a text change and it has been reviewed and approved by Dan.

PR #319, opened 2023-01-31
Added back autosave on save failures [#183842099]

dougmartin:
This ensures when a network error resolves itself autosave continues to run.


lublagg (approved):
Looks good 👍

PR #320, opened 2023-02-01
Fixed double click of top level Google drive list [#184360746]

dougmartin:
This also adds a "Loading..." message which was found to be needed for large file lists.


lublagg (approved):
Looks good :)

PR #322, opened 2023-02-02
fix: :bug: Removed changed client provider check in file dialog tab

dougmartin:
This prevented a file opened in the readonly provider from being saved in the google provider


lublagg (approved):
👍

PR #323, opened 2023-02-07
Korean

eireland:
Adds Korean translation strings


dougmartin:
@eireland This shouldn't be based on master as there are a lot of commits on master that are not ready to be deployed to production (still in QA).  Instead they should be based on the v1.5.13 tag (the latest release).

dougmartin:
Also we should only add the new Korean language file as updating the others may cause a merge conflict with what is merged into master right now.  If you want I can update the PR for you.

dougmartin:
Looks good :+1:

PR #325, opened 2023-02-16
Fixed missing "Loading..." message [#184485531]

dougmartin:
This also fixes a css color issue on the clear "icon" so that it is visible in CODAP.


lublagg (approved):
Looks good :)

PR #326, opened 2023-02-16
Remove extension from local file save [#184486444]

dougmartin:
This fixes a recent change made when adding extensions to the export dialog.  The export dialog still has the extension but the local file save no longer does (as requested).


lublagg (approved):
👍

PR #327, opened 2023-02-16
Added alert popup when Google file parameter is detected [#184485563]

dougmartin:
This should cause the browser popup blocker to not block the popup as it is handled via a user gesture


lublagg (approved):
Looks good 👍

PR #330, opened 2023-03-05
fix: Shortcuts to folders in My Drive appear as regular files (PT-184617375)

dougmartin:
This fixes both listing shortcuts to folders as folders and saving into folder shortcuts.


pjanik (approved):
Looks good 👍

Issue #112, opened 2018-11-16
Can't build CFM using Node v10


pjanik:
Perhaps some outdated dependency. Error:
```
$ gulp default

gulp[25840]: ../src/node_contextify.cc:627:static void node::contextify::ContextifyScript::New(const FunctionCallbackInfo<v8::Value> &): Assertion `args[1]->IsString()' failed.
 1: 0x100038c65 node::Abort() [/Users/piotr/.nvm/versions/node/v10.9.0/bin/node]
 2: 0x100037d5f node::MakeCallback(v8::Isolate*, v8::Local<v8::Object>, char const*, int, v8::Local<v8::Value>*, node::async_context) [/Users/piotr/.nvm/versions/node/v10.9.0/bin/node]
 3: 0x1000606f5 node::contextify::ContextifyScript::New(v8::FunctionCallbackInfo<v8::Value> const&) [/Users/piotr/.nvm/versions/node/v10.9.0/bin/node]
 4: 0x10023c58f v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo*) [/Users/piotr/.nvm/versions/node/v10.9.0/bin/node]
 5: 0x10023b72b v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<true>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) [/Users/piotr/.nvm/versions/node/v10.9.0/bin/node]
 6: 0x10023b147 v8::internal::Builtin_Impl_HandleApiCall(v8::internal::BuiltinArguments, v8::internal::Isolate*) [/Users/piotr/.nvm/versions/node/v10.9.0/bin/node]
 7: 0x3d777b55c01d
Abort trap: 6
```
Node v8 seem to work fine.

dougmartin:
I hit the same issue when I was setting up Travis so I forced it to Node 8:

https://github.com/concord-consortium/cloud-file-manager/blob/master/.travis.yml#L4

PR #333, opened 2023-05-04
chore: Add "engines" to package.json (fixes GH-112 & PT-185071892)

433eros:
Only allow using Node.js `>= 16` since we use Node.js v16 with Travis: https://github.com/concord-consortium/cloud-file-manager/blob/1daeff6cdf248e0b379328704748b0862e623898/.travis.yml#L4-L5


dougmartin (approved):
Looks good :+1:

PR #336, opened 2023-07-18
feat: Added content load filter [PT-185485076]

dougmartin:
This will allow SageModeler to rewrite the iframe title that was broken with the early Korean translations.


pjanik (approved):
Looks good 👍

PR #338, opened 2023-07-26
fix: Missing broadcast of logLaraData from interactive-api-provider [PT-184770212]

dougmartin:
This was causing CODAP to not log the run_remote_endpoint when it logged messages.


dougmartin:
Deleting to clean up my old lingering PRs.

PR #340, opened 2023-11-29
chore: update dependencies; fix lint issues; general clean up

kswenson:
- update dependencies (`webpack` 5, `lara-interactive-api`, `token-service`, etc.)
- replace node-only `mime-type` library with node/browser `mime` library
- fix lint warnings (mostly semicolons)
- update jest tests to eliminate warnings
- fix `client-logging-provider` example
- remove `Enable Interactive API sharing` example from list since it doesn't exist 😬 
- separate `CloudFileManager` into its own module (out of `app.js`)


src/code/client.ts
```diff
@@ -201,7 +201,7 @@ class CloudFileManagerClient {
         if (allProviders[providerName]) {
           Provider = allProviders[providerName]
           // don't add providers that require configuration if no (or invalid) configuration provided
-          if (Provider.hasValidOptions(providerOptions)) {
+          if (!Provider.hasValidOptions || Provider.hasValidOptions(providerOptions)) {
```
dougmartin, line 204: Did you test this with all the providers?

dougmartin (approved):
Looks good :+1:

src/code/client.ts
kswenson: Yes, all the other providers have an implementation for `hasValidOptions` so they should be unaffected by the change.

PR #346, opened 2023-12-19
Changed Google project to use "CFM Examples" project

dougmartin:
Also removed the google-picker-dev.html example as that is no longer needed.


eireland (approved):
👍 LGTM

PR #347, opened 2023-12-19
fix: Remove search query on Google picker file opens

dougmartin:
Automatically setting the query obscures the folders.


eireland (approved):
👍 LGTM

PR #349, opened 2024-01-10
feat: support publishing CFM as npm module

kswenson:
The basic approach here follows a LogRocket post on [Publishing Node modules with TypeScript and ES modules](https://blog.logrocket.com/publishing-node-modules-typescript-es-modules/).

We leave the webpack configuration alone so that the existing builds used by CODAP and Sage should be unaffected. Instead, the TypeScript compiler, `tsc`, is used to generate additional folders in the `dist` folder, and then the entire `dist` folder is published. Publishing this way creates individual source files in the `dist/cjs` and `dist/esm` folders rather than bundles, but this is fine given that clients (e.g. CODAP v3) will generally be bundling the imported code anyway.

Nothing has been published to npm yet, but I have used [yalc](https://www.npmjs.com/package/yalc) to publish locally and then imported the locally published cfm into CODAP v3 and demonstrated that things work as expected to a first approximation.

Noteworthy pieces of this PR are:
1. Rename module in `package.json` to `@concord-consortium/cloud-file-manager` so it will be published as a scoped package.
2. Add `package.json` fields to support publication: `main`, `module`, `types`, `style`, `files`.
3. Add new scripts to support publication: `build:all`, `build:library`, `prepublishOnly`.
4. Add new scripts to support various publish options: `publish:npm:latest`, `publish:npm:latest:preview`, `publish:npm:next`, `publish:npm:next:preview`, `publish:yalc`.
5. Move `react` and `react-dom` from `dependencies` to `peerDependencies` so that clients can bring their own version of React (e.g. React 18).
6. Support client use of React 18 by adding a `renderRoot` option to CFM configuration so that clients can make use of the `ReactDOM.createRoot` API. There's a new React 18 example which demonstrates its use.
7. Require explicit import of jquery. There are six libraries bundled into the `globals` module. All of them are accessed via global variables and webpack is configured to convert all references to those modules into references via those global variables. For unknown reasons, the code was explicitly importing five of them, but jquery was just assumed to be present globally without being explicitly imported. Because of the webpack configuration, adding the explicit import doesn't change the generated code, but it makes it easier to import the code into another client (e.g. CODAP v3) which doesn't require the same global variable configuration.
8. Fix some CSS `url()` paths that were incorrect and which failed when imported into CODAP v3. It's not clear to me why these incorrect paths seemed to be working before, but I've tested them with the existing examples and they still work after the changes. 🤷 
9. Update the CFM version number to `2.0.0-pre.1` -- seems like a reasonable time to declare a 2.0 release, what with the recent google drive changes as well as the new publication option.


dougmartin (approved):
Looks good :+1: 

I agree using ESM modules makes the most sense instead of bundling with something like rollup since only more modern apps will be requiring this.

I tested the branch with SageModeler and it seems to work well:

https://sagemodeler.concord.org/app/?cfmBaseUrl=npm-module-tsc&sage:cfmBaseUrl=npm-module-tsc&codap:cfmBaseUrl=npm-module-tsc

PR #350, opened 2024-01-11
chore: update comments and ReadMe; eliminate more semicolons for consistency

kswenson:
This started with the simple task of updating the ReadMe with information about the new `@concord-consortium/cloud-file-manager` library option.
- While at it I decided to add some information about the options that could be passed to the CFM in the ReadMe as well.
- While doing that it seemed like some additional comments about the meaning of some of the options was warranted, although I bailed out before documenting all of them.
- Finally, while looking at the definitions of the various options I was struck by the number of semicolons in this supposedly semicolon-less project and updated the ESLint configuration to make things more consistent.


dougmartin (approved):
Looks good :+1:

PR #353, opened 2024-01-19
fix: Remove select different account link for now [PT-186862486]

kswenson:
fix: Remove select different account link for now [PT-186862486]
(cherry picked from commit f253f6192b832021bf0b65ba47b82b133044a0bf)


kswenson:
Merging without review as it's just applying the previously reviewed hot-fix from the 1.9.1 branch to keep `master` in sync.

PR #354, opened 2024-01-26
feat: improve ability to import as npm library (support CODAP v3)

kswenson:
The goal here, as with other recent PRs, is to enable/improve support for clients such as CODAP v3 to load the CFM as an npm library while not changing the behavior of the CFM for existing clients such as CODAP v2 and Sage Modeler.

- New configuration option `isClientContent`
  - Previously, CFM assumed that client content had a top-level `metadata` property. This assumption is preserved for backward-compatibility, but now clients can specify a custom function for identifying client content. CODAP v3 uses this to pass a custom function that can recognize CODAP v2 or v3 documents.
- Fix `ProviderInterface.getClientContent` so that it respects the `wrapFileContent` setting.
  - The `getClientContent` function is one side of a coin with the `getContent` function on the other side.
  - The `getContent` function already made use of the `wrapFileContent` setting to determine whether to return top-level or wrapped content.
  - The `getClientContent` function in contrast made an assumption about the form of the content rather than respecting the `wrapFileContent` setting.
  - The CODAP v3 file format has a `content` property which violates the content assumption made by the previous code, but use of the `wrapFileContent` setting should lead to the correct result in all circumstances.
-  Use enumerated `CloudFileManagerEventType` rather than `string` for CFM event types for better type-safety in event manipulation code.
- Fix latent race condition in which clients could call into CFM before necessary event handlers were installed.
  - If the client calls the `clientConnect` function promptly the CFM can send the `connected` event to the client before the CFM has been fully initialized. In particular, event handlers for functions like `setMenuBarInfo` are installed in `AppView`'s `componentDidMount` method. If the client calls `setMenuBarInfo` upon receipt of the `connected` event, this can occur before the necessary handlers have been installed.
  - Therefore, we add a promise which resolves when initialization is complete and wait for that promise to be resolved before handling any client callbacks.
- Fix styling of top-level `view` class.
  - When iframed, CFM's top-level `div` is given a class of `app`.
  - When not iframed, CFM's top-level `div` is given a class of `view`.
  - TODO for the future: use less generic class names to avoid conflicts with clients.
  - The CFM .css explicitly styled the `app` class, but relied on clients to style the `view` class.
  - The primary client of non-iframed CFM is CODAP, which styled the `view` class identically to the `app` class, namely absolutely positioned and taking up 100% of its parent.
  - Therefore, we move that styling into the CFM's .css for consistency.
- Fix implementation of stylus `absolute` function
  - In debugging the previous .css change, I noticed invalid `width` and `height` properties being generated for the `app` and `view` classes.
  - This turned out to be because the `absolute` function defaulted all of its arguments to `'auto'`. Unfortunately, while `auto` is a valid value for `height` or `width`, `'auto'` is not.
  - It turns out that all existing calls to the `absolute` function specified exactly four arguments -- `top`, `left`, `bottom`, `right` -- and so the resulting `height` or `width` values were always invalid and hence always ignored by the browser.
  - Therefore, the `absolute` function has been changed to require exactly four arguments and not perform any argument-defaulting.

### DIFF
--- a/src/code/app-options.ts
+++ b/src/code/app-options.ts
@@ -132,6 +132,7 @@ export interface CFMAppOptions {
   // false (default) if CFM sets page title from document name
   appSetsWindowTitle?: boolean
   wrapFileContent?: boolean
+  isClientContent?: (content: unknown) => boolean
   mimeType?: string
   // note different capitalization from CFMBaseProviderOptions
   readableMimeTypes?: string[]
--- a/src/code/client.test.ts
+++ b/src/code/client.test.ts
@@ -14,9 +14,9 @@ mockApi.getInitInteractiveMessage.mockImplementation(() => Promise.resolve({
 describe("CloudFileManagerClientEvent", () => {
 
   test('increments id with each event registered', () => {
-    const clientEvent = new CloudFileManagerClientEvent("any")
+    const clientEvent = new CloudFileManagerClientEvent("connected")
     expect(clientEvent.id).toBe(1)
-    const clientEvent2 = new CloudFileManagerClientEvent("any")
+    const clientEvent2 = new CloudFileManagerClientEvent("ready")
     expect(clientEvent2.id).toBe(2)
   })
 })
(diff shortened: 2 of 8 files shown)


dougmartin (approved):
Looks good :+1: 

I like the typing of the events.  We should have done that in the initial "decaffination" process.

PR #357, opened 2024-02-21
fix: export CFM version number rather than `__PACKAGE_VERSION__`

kswenson:
The CFM code contains the string `__PACKAGE_VERSION__` in several places. The expectation is that this string will be replaced by the actual CFM version number from `package.json` at build time. For the webpack build, this is handled by the `ReplaceInFileWebpackPlugin`. For the library export we don't have a bundler that can handle the string replacement. Therefore, this PR introduces the `build-library.sh` bash script which builds the exported library folders with calls to `tsc`, as before, then also uses `grep` and `sed` to perform the string replacement of the version number. The script was developed with the help of ChatGPT.

### DIFF
--- a/bin/build-library.sh
+++ b/bin/build-library.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Developed with the help of ChatGPT
+
+# Set the locale for the script to the C locale
+export LC_ALL=C
+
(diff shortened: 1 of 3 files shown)


eireland (approved):
👍 LGTM

PR #359, opened 2024-10-20
feat: support dynamic language switching

kswenson:
CODAP v2 required a full page reload to switch language due to SproutCore. CODAP v3 will support dynamic language switching without a page reload, so the CFM needs to as well. With this PR, CFM supports the notion of current language rather than just a default language set on startup. The translation function and the rendering of the menus respect the current language. Since the File menu is provided by the client, there's a new API for replacing the existing menu.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,10 +2,10 @@
   "name": "@concord-consortium/cloud-file-manager",
   "description": "Wrapper for providing file management for web applications",
   "author": "The Concord Consortium",
-  "version": "2.0.0-pre.4",
+  "version": "2.0.0-pre.6",
   "repository": {
     "type": "git",
-    "url": "https://github.com/concord-consortium/cloud-file-manager"
+    "url": "git+https://github.com/concord-consortium/cloud-file-manager.git"
   },
   "engines": {
     "node": ">= 16",
@@ -107,7 +107,7 @@
     "publish:npm:next": "npm publish --access public --tag next",
     "publish:npm:next:preview": "npm publish --access public --tag next --dry-run",
     "publish:yalc": "npx yalc publish",
-    "strings:build": "./node_modules/.bin/strip-json-comments src/code/utils/lang/en-US-master.json > src/code/utils/lang/en-US.json",
+    "strings:build": "strip-json-comments src/code/utils/lang/en-US-master.json > src/code/utils/lang/en-US.json",
     "strings:pull:usage": "echo Usage: `npm run strings:pull -- -a <poeditor_api_key>`",
     "strings:pull": "./bin/strings-pull-project.sh",
     "strings:push:usage": "echo Usage: `npm run strings:push -- -a <poeditor_api_key>`",
--- a/readme.md
+++ b/readme.md
@@ -138,8 +138,8 @@ export interface CFMMenuBarOptions {
   languageMenu?: {
     currentLang: string
     options: { label: string, langCode: string }[]
+    onLangChanged?: (langCode: string) => void
   }
-  onLangChanged?: (langCode: string) => void
 }
 
 export interface CFMShareDialogSettings {
--- a/src/code/app-options.ts
+++ b/src/code/app-options.ts
@@ -19,8 +19,8 @@ export interface CFMMenuBarOptions {
   languageMenu?: {
     currentLang: string
     options: { label: string, langCode: string }[]
+    onLangChanged?: (langCode: string) => void
   }
-  onLangChanged?: (langCode: string) => void
 }
 
 export interface CFMShareDialogSettings {
@@ -28,12 +28,15 @@ export interface CFMShareDialogSettings {
   serverUrlLabel?: string
 }
 
-export interface CFMUIOptions {
-  menuBar?: CFMMenuBarOptions
+export interface CFMUIMenuOptions {
   // null => no menu; undefined => default menu
   menu?: CFMMenu | null
   // map from menu item string to menu display name for string-only menu items
   menuNames?: Record<string, string>
+}
+
+export interface CFMUIOptions extends CFMUIMenuOptions {
+  menuBar?: CFMMenuBarOptions
   // used for setting the page title from the document name (see appSetsWindowTitle)
   windowTitleSuffix?: string
   windowTitleSeparator?: string
--- a/src/code/ui.ts
+++ b/src/code/ui.ts
@@ -9,7 +9,7 @@
  * Full docs: https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md
  */
 import { CloudFileManagerClient } from './client'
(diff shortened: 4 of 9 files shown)


package.json
```diff
@@ -2,10 +2,10 @@
   "name": "@concord-consortium/cloud-file-manager",
   "description": "Wrapper for providing file management for web applications",
   "author": "The Concord Consortium",
-  "version": "2.0.0-pre.4",
+  "version": "2.0.0-pre.6",
```
kswenson, line 5: Published `pre.5` during testing.

package.json
```diff
@@ -2,10 +2,10 @@
...
   "repository": {
     "type": "git",
-    "url": "https://github.com/concord-consortium/cloud-file-manager"
+    "url": "git+https://github.com/concord-consortium/cloud-file-manager.git"
```
kswenson, line 8: Recommended during `npm publish`.

package.json
```diff
@@ -107,7 +107,7 @@
     "publish:npm:next": "npm publish --access public --tag next",
     "publish:npm:next:preview": "npm publish --access public --tag next --dry-run",
     "publish:yalc": "npx yalc publish",
-    "strings:build": "./node_modules/.bin/strip-json-comments src/code/utils/lang/en-US-master.json > src/code/utils/lang/en-US.json",
+    "strings:build": "strip-json-comments src/code/utils/lang/en-US-master.json > src/code/utils/lang/en-US.json",
```
kswenson, line 110: Recommended during `npm publish`.

readme.md
```diff
@@ -138,8 +138,8 @@ export interface CFMMenuBarOptions {
   languageMenu?: {
     currentLang: string
     options: { label: string, langCode: string }[]
+    onLangChanged?: (langCode: string) => void
```
kswenson, line 141: Fixes the TypeScript type to reflect the pre-existing reality.

src/code/app-options.ts
```diff
@@ -19,21 +19,24 @@ export interface CFMMenuBarOptions {
   languageMenu?: {
     currentLang: string
     options: { label: string, langCode: string }[]
+    onLangChanged?: (langCode: string) => void
```
kswenson, line 22: Fixes the TypeScript type to reflect the pre-existing reality.

src/code/app-options.ts
```diff
@@ -19,21 +19,24 @@ export interface CFMMenuBarOptions {
...
 }
 
 export interface CFMShareDialogSettings {
   serverUrl?: string
   serverUrlLabel?: string
 }
 
-export interface CFMUIOptions {
-  menuBar?: CFMMenuBarOptions
+export interface CFMUIMenuOptions {
```
kswenson, line 31: Pulled these out into a base type to support the new `replaceMenu` API.

src/code/ui.ts
```diff
@@ -97,7 +97,7 @@ class CloudFileManagerUIMenu {
       const item = menuItems[i]
       if (item === 'separator') {
         menuItem = {
-          key: `seperator${i}`,
+          key: `separator${i}`,
```
kswenson, line 100: Fixing an unrelated typo.

src/code/utils/translate.ts
```diff
@@ -84,23 +84,33 @@ languageFiles.forEach(function(lang) {
...
-  if (lang == null) { lang = defaultLang }
+  if (lang == null) { lang = gCurrentLanguage }
   lang = lang.toLowerCase()
   let translation = translations[lang] != null ? translations[lang][key] : undefined
   if ((translation == null)) { translation = key }
   return translation.replace(varRegExp, function(match: string, key: string) {
     return Object.prototype.hasOwnProperty.call(vars, key)
             ? vars[key]
-            : `'** UKNOWN KEY: ${key} **`
+            : `'** UNKNOWN KEY: ${key} **`
```
kswenson, line 113: Fixing an unrelated typo.

src/code/utils/translate.ts
```diff
@@ -32,22 +32,22 @@ interface LanguageFileEntry {
...
+  {key: 'he',       contents: he},     // Hebrew
+  {key: 'ja' ,      contents: ja},     // Japanese
+  {key: 'ko' ,      contents: ko},     // Korean
+  {key: 'nb',       contents: nb},     // Norwegian Bokmål
+  {key: 'nn',       contents: nn},     // Norwegian Nynorsk
+  {key: 'pl',       contents: pl},     // Polish Polski
+  {key: 'pt-BR',    contents: ptBR},   // Brazilian Portuguese
+  {key: 'th',       contents: th},     // Thai
+  {key: 'tr',       contents: tr},     // Turkish
+  {key: 'zh-Hans',  contents: zhHans}, // Simplified Chinese
```
dougmartin, line 49: There is something in the back of my brain that tells me we need to keep it just `zh`.  I think it is because we want any Chinese variant other than Taiwanese to default to Hans which it will do if it is just `zh`.  We don't have to do the same for `en-GB`/`en-CA` to use `en-US` etc because `en` is the default language used if the specific language requested by the browser is not used.

dougmartin (commented):
I think we need to set `zh-Hans` back to just `zh` but you might want to test if that is true by setting your browser to `zh-CN` or `zh-HK` and seeing if you see Chinese or English.

src/code/utils/translate.ts
kswenson: I'll see if I can figure that out. Other things weren't working with switching due to the code mismatch between CODAP and CFM, so if you're right that `zh` needs special treatment then we may require some special-case code.

kswenson: I'll do the experiment, but I think it should be fine because both CODAP and CFM use the following code to add languages to the map:
```typescript
const translations: Record<string, LanguageFileContent> =  {}
languageFiles.forEach(function(lang) {
  translations[lang.key.toLowerCase()] = lang.contents
  // accept full key with region code or just the language code
  const baseLang = getBaseLanguage(lang.key)
  if (baseLang && !translations[baseLang]) {
    translations[baseLang] = lang.contents
  }
})
```
so `zh-Hans` will be added to the map under both `zh-Hans` and `zh` because the code specifies that in the case of multiple instances of the same language, the first one takes precedence.

kswenson: I just ran the experiment. If I set my preferred browser language to simply Chinese, I get `zh-Hans`.

dougmartin: I remember what it was - this was something in SageModeler when it passed down the language to CODAP in the url.  We had to only give it the major language code, I think because CODAP only had `zh`. 

https://github.com/concord-consortium/sage-modeler-site/blob/master/src/code/codap-iframe-src.ts#L20-L24

So I guess no change is needed.

dougmartin (approved):
Looks good :+1:

PR #361, opened 2024-11-12
add openLocalFileWithConfirmation

scytacki:
This checks if the document is dirty and asks for confirmation if it is. Then it opens the passed in file object.

Other changes:
- example-app is updated to test this
- s3-share-provider is removed all-providers.html to prevent warning about it not being valid
- typo fixed in s3-provider comment
- npm yalc commands updated to match other projects
- readme updated with yalc instructions
- new providers.md started to document the providers


scytacki:
@kswenson one thing I don't have my head around is how the CFM works when CODAP is embedded in an iframe the way SageModeler uses it. I think in that case many of these client functions are "proxied" over postMessage to the parent frame. In that case this new function won't work. Probably it is fine to not deal with that now, but I wanted to point out that it isn't something I've figured out or tested.

kswenson (approved):
👍 LGTM

kswenson:
> @kswenson one thing I don't have my head around is how the CFM works when CODAP is embedded in an iframe the way SageModeler uses it. I think in that case many of these client functions are "proxied" over postMessage to the parent frame. In that case this new function won't work. Probably it is fine to not deal with that now, but I wanted to point out that it isn't something I've figured out or tested.

It's not an area I'm very familiar with, either. @dougmartin may have a better idea.

PR #366, opened 2025-03-12
add share button to no-frame example

scytacki:
This fixes a bug when sharing plain text content.
It also fixes a bug that was breaking all sharing in these examples because the sharing provider uses at `process.env`.

The `process.env` reference still exists in the library, so any application using the CFM library needs to handle this reference on its own.

TODO:
- [x] change the token service environment to staging for the examples, so they don't polute the production CFM.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,6 +52,7 @@
         "cypress": "^7.7.0",
         "cypress-commands": "^1.1.0",
         "cypress-xpath": "^1.6.2",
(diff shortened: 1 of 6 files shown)


scytacki:
This is deployed here:
https://models-resources.concord.org/cloud-file-manager/branch/share-example/examples/no-frame.html

dougmartin (approved):
Looks good :+1:

PR #378, opened 2025-08-25
French dutch language updates

ddamelin:
this has tweaks to the French and Dutch translations of CFM


dougmartin (approved):
Looks good :+1:

PR #418, opened 2026-03-20
CODAP-1198: fix Google Picker hang caused by React Aria Modal inert

kswenson:
## Summary

- React Aria Modal sets `inert` on all elements outside the modal scope, which makes the Google Picker overlay completely unresponsive when opened from the file dialog
- This is a [known React Aria issue](https://github.com/adobe/react-spectrum/issues/8784) with an [in-progress fix](https://github.com/adobe/react-spectrum/pull/8796) not yet released
- Work around it with a MutationObserver that catches Google Picker elements as they are added to the DOM and marks them with `data-react-aria-top-layer` so React Aria treats them as part of the modal stack
- Bumps version to 2.2.6

## Test plan

- [ ] Open the file dialog via keyboard, navigate to Google Drive tab, verify the picker appears and is interactive (clicks work, files can be selected)
- [ ] Open the file dialog via mouse, verify Google Drive picker still works as before
- [ ] After dismissing the picker (cancel), reopen it and verify it still works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@concord-consortium/cloud-file-manager",
-  "version": "2.2.5",
+  "version": "2.2.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@concord-consortium/cloud-file-manager",
-      "version": "2.2.5",
+      "version": "2.2.6",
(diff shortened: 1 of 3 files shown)


emcelroy (approved):
Looks good 👍

